PR #52, opened 2024-04-05
Fix library types

ivy-lli:
We need to use a thridparty utility to generate d.ts files, because typescript does not resolve path aliases.

button.d.ts with tsc:
```ts
import { Flex, type IvyIconProps } from '@/components/common';
```

button.d.ts with vite-plugin-dts:
```ts
import { Flex, IvyIconProps } from '..';
```

'@' is an tsconfic path alias for './src' and this can not be resolved if the package is a dependency in the node_modules folder.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3260,6 +3260,85 @@
         "react": ">=16"
       }
     },
+    "node_modules/@microsoft/api-extractor": {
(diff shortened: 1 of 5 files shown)


ivy-cst:
:open_mouth:

PR #74, opened 2024-05-08
Fix theme without localstorage usage

ivy-lli:


packages/components/src/context/useTheme.tsx
```diff
@@ -25,11 +25,11 @@ const ThemeProviderContext = createContext<ThemeProviderState>(initialState);
 export const ThemeProvider = ({
   children,
   defaultTheme = 'system',
-  storageKey = 'ivy-ui-theme',
+  storageKey,
```
ivy-lli, line 28: I think this is the problem for https://1ivy.atlassian.net/browse/XIVY-14090
As soon as you once change the theme in the editor, the localstorage was filled with the 'ivy-ui-theme' property which holds the theme, after that the defaultTheme parameter is no longer respected (e.g. the in the eclipse integration -> theme URL parameter).

Now the localstorage is only used if this parameter is defined.

ivy-rew (approved):

PR #87, opened 2024-06-07
Add dialog component

ivy-lli:
the base for a browser component, but a dialog is needed anyways
FYI @ivy-fhe


ivy-cst (approved):
Looks good :+1: 
For me it looks like the dialog is like a complete component. May be the base of a browser and dialog is another common overlay component?

PR #88, opened 2024-06-07
Add tabs component

ivy-lli:


packages/components/src/components/common/tabs/tabs.stories.tsx
```diff
@@ -0,0 +1,38 @@
...
+};
+
+export default meta;
+
+type Story = StoryObj<typeof Tabs>;
+
+export const Default: Story = {
+  render: props => (
+    <Tabs defaultValue='attribute' className='w-[400px]' {...props}>
+      <Flex direction={props.orientation === 'horizontal' ? 'column' : 'row'} gap={4}>
```
ivy-cst, line 22: Why is this flex needed? If I look here this is not needed: https://www.radix-ui.com/primitives/docs/components/tabs

ivy-cst (approved):
Looks good :+1:

packages/components/src/components/common/tabs/tabs.stories.tsx
ivy-lli: to put some space between the tabs and the content :D
Only to make the story looks a little bit better. can also use plain CSS like the radix-ui doc does it ;)

PR #90, opened 2024-06-11
Fix expand tree button with select rows

ivy-lli:
Before (first only changes selection of row, second will trigger the collapse/expand):
Screen Recording 2024-06-11 at 08 15 36

After fix (expand will no longer trigger the selection):
Screen Recording 2024-06-11 at 08 12 18

IMO good behavior and anyway better that the the collapse/expand only trigger on the second click. Maybe you don't want to change the selection with a click on the expand button...

### DIFF
--- a/packages/components/src/components/common/table/edit/edit.tsx
+++ b/packages/components/src/components/common/table/edit/edit.tsx
@@ -1,4 +1,4 @@
-import { Input, type InputProps, SimpleSelect, type SimpleSelectProps } from '@/components';
+import { Input, type InputProps, BasicSelect, type BasicSelectProps } from '@/components';
 import type { CellContext, RowData } from '@tanstack/react-table';
 import * as React from 'react';
 import { editCell } from './edit.css';
@@ -36,13 +36,13 @@ const InputCell = <TData,>({ cell, className, ...props }: InputCellProps<TData>)
 };
 InputCell.displayName = 'InputCell';
 
-type SelectCellProps<TData> = SimpleSelectProps & {
+type SelectCellProps<TData> = BasicSelectProps & {
   cell: CellContext<TData, string>;
 };
 
 const SelectCell = <TData,>({ cell, className, ...props }: SelectCellProps<TData>) => {
   const { value, updateValue, className: editCell } = useEditCell(cell);
-  return <SimpleSelect value={value} onValueChange={updateValue} className={cn(editCell, className)} {...props} />;
+  return <BasicSelect value={value} onValueChange={updateValue} className={cn(editCell, className)} {...props} />;
 };
 SelectCell.displayName = 'SelectCell';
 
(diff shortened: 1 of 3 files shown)


packages/components/src/components/common/table/table.css.ts
```diff
@@ -36,15 +35,12 @@ export const footer = style({
   backgroundColor: vars.color.n100
 });
 
-export const row = style([
-  transitionColors,
```
ivy-lli, line 40: removed also the transition colors on rows (so the select color has an animation, see gifs).
Because it also has strange side effects, e.g. on trees and IMO not that important :)

packages/components/src/components/common/table/edit/edit.tsx
```diff
@@ -36,13 +36,13 @@ const InputCell = <TData,>({ cell, className, ...props }: InputCellProps<TData>)
...
 
-type SelectCellProps<TData> = SimpleSelectProps & {
+type SelectCellProps<TData> = BasicSelectProps & {
   cell: CellContext<TData, string>;
 };
 
 const SelectCell = <TData,>({ cell, className, ...props }: SelectCellProps<TData>) => {
   const { value, updateValue, className: editCell } = useEditCell(cell);
-  return <SimpleSelect value={value} onValueChange={updateValue} className={cn(editCell, className)} {...props} />;
+  return <BasicSelect value={value} onValueChange={updateValue} className={cn(editCell, className)} {...props} />;
```
ivy-lli, line 45: changed also to BasicSelect (SimpleSelect is deprecated as I streamlined all basic component implementations with a Basic prefix)

packages/components/src/components/common/table/tree/tree.tsx
```diff
@@ -21,7 +21,10 @@ const expanedButton = <TData,>(row: Row<TData>, lazy?: LazyExpand<TData>) => {
         className={expandButton}
         aria-label={row.getIsExpanded() ? 'Collapse row' : 'Expand row'}
         data-state={row.getIsExpanded() ? 'expanded' : 'collapsed'}
-        onClick={row.getToggleExpandedHandler()}
+        onMouseDown={row.getToggleExpandedHandler()}
+        onKeyDown={e => {
+          if (e.key === 'Enter' || e.key === ' ') row.getToggleExpandedHandler()();
+        }}
```
ivy-lli, line 27: removed the onclick, because this also triggers first onFocus, which selects the row (if it is in a SelectRow), but somehow not the click itself...
So, I changed it to an onMouseDown / onKeyDown event listener to toggle the expansion.

ivy-cst (commented):
Looks nice :+1:

ivy-lgi (commented):
thx :+1:

ivy-lgi:
FYI @ivy-fhe

PR #91, opened 2024-06-11
Add browser component

ivy-lli:
Restructure storybook a bit (common component section, editor section with more detailed components)

Browser:
- First draft
- Multiple browsers, switch per tabs
- autofocus on the search field
- Tree table with select node, apply selection, etc.
- Support to lazy load info text
- Open it inside of a dialog or somewhere else (maybe a popover is the better choice in the end?)

Open points:
- Maybe a table is also needed without the need to define empty child, not tried it out...
- Lazy load tree children's (needed for e.g. attribute browser)
- Maybe an additional attribute for the inserted value? currently, the value will be needed which is also shown as the main name of the tree node (but maybe this is not the right one, and an additional name prop is needed on the BrowserNode?)
- Not sure if the load data on search is already supported
- Check performance (we currently use visualizer libs and only show the first 100 entries techniques in the inscription browsers). I think the storybook needs to have a mode where we show a lot of data, e.g. load it from a real engine to test the real world as much as possible. or cleverly generate stuff on the fly...

Not sure if I covered all points, but I think it is a good first draft and we can try to use it in the inscription and also on other points to test if other parts are missing...


ivy-cst (approved):
Looks good :+1: 
Desk review?

ivy-edp (approved):
Nice job on the first draft! 👍 Maybe a short GIF would have helped me to even better understand what's already functional.

Regarding the additional attribute: I agree that a solution is needed since the inserted value in the CMS, Function, and Type-browser varies depending on factors like location and type.

I thinks this should already work, but just to make sure can you already define specific icons for each cell? For instance, the CMS-Browser currently defines icons based on its type.

Additionally, two features from the type-browser are currently probably not included:

1. The ability to add a predefined search term. For example, if you select text and open the browser, the selected text should be inserted into the search bar.
2. Feedback when no data is found for the entered search term. In the Typebrowser if there is no type found there is a short infomation that the entered search term has no fitting type.

Yes, I also think the best approach for this generic browser is to start implementing these in the inscription and then test to see what works and what doesn't.

ivy-lli:
why a gif, when you can play live with it ;) https://jenkins.ivyteam.io/job/ui-components/job/browser/

1. icons: currently you can define the icons of each single BrowserNode 
2. init search filter: yes true, this is currently missing
3. no data feedback: also true, but I think currently the filter is anyway strange and not really working. Need to have a look on that. Maybe we need to have some default filter behaviour from which a dev can choose from (e.g. the one @ivy-lgi implemented for the variable editor)

PR #94, opened 2024-06-12
XIVY-14003 fix indentation in tree

ivy-lgi:
On root level, icons are not aligned if some of the entries are collapsible, while they are aligned once the depth is bigger than 0. I think it should look the same in both cases. Thoughts?

Before:
image

After:
image


ivy-lli (approved):
yes I think you're right :)

PR #97, opened 2024-06-12
XIVY-14003 adjust icon sizes

ivy-lgi:
Anaïs said to make all the icons in the variables editor the same size (16px).
I think this should be a global change.
Maybe we should think about streamlining the size of all icons in general to 16px?

### DIFF
--- a/packages/components/src/components/common/table/tree/tree.css.ts
+++ b/packages/components/src/components/common/table/tree/tree.css.ts
@@ -8,3 +8,7 @@ export const expandButton = style({
     }
   }
 });
+
+export const cellIcon = style({
+  fontSize: '16px'
+});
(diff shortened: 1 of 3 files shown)


packages/components/src/components/common/table/tree/tree.css.ts
```diff
@@ -8,3 +8,7 @@ export const expandButton = style({
     }
   }
 });
+
+export const cellIcon = style({
+  fontSize: '16px'
```
ivy-lli, line 13: maybe 15px would be better, as then the rows would stay at a height of 32px. With 16px they grow to 33px.
But not sure if this is really a problem :)
So I think this change is fine for me 👍

ivy-lli (approved):

PR #98, opened 2024-06-14
XIVY-14003 new icons

ivy-lgi:
icon for drag and drop reorder in table
icon for empty details view of variables editor

Old:
image

New:
image


packages/icons/src/icons/edit-dots copy.svg
weissreto, line 1: Does this file need to be named like this (copy)? 
Shouldn't it be edit-dots.svg

ivy-lgi: my bad, the icon was already present :sweat_smile:

packages/icons/src/icons/drag-drop.svg
weissreto, line 1: Why add this icon if we don't use it anywhere?

weissreto (approved):
Another small question. Otherwise good.

packages/icons/src/icons/drag-drop.svg
ivy-lgi: the variables editor gets its icons from here, it will be used there.

ivy-lgi: See https://github.com/axonivy/config-editor-client/pull/109

PR #109, opened 2024-06-25
Remove z-index form default dialog styles

ivy-lli:
- Portal should handle that the dialog is on top (otherwise set z-index manual)

image


ivy-lmu (approved):
thx 👍

PR #111, opened 2024-07-03
XIVY-13261 Add ToggleBar Component

ivy-edp:
My first component for the UI components. I would then like to use the ToggleBar in the form editor. Feedback?
I wasn't quite sure whether I should implement the ToggleBarItemOnlyIcon component as I did or simply work with the ToggleBarItem component and filter out this scenario with if-statements. What do you think?

grafik

### DIFF
--- a/packages/components/src/components/common/toggleGroup/toggleGroup.css.ts
+++ b/packages/components/src/components/common/toggleGroup/toggleGroup.css.ts
@@ -1,0 +1,21 @@
+import { vars } from '@/styles/theme.css';
+import { recipe } from '@vanilla-extract/recipes';
+
+export const toggleGroup = recipe({
+  base: {
+    display: 'inline-flex',
+    alignItems: 'center',
+    gap: vars.size.s2,
+    padding: '10px',
+    backgroundColor: vars.color.background,
+    borderRadius: vars.border.r2
+  },
+  variants: {
+    gap: {
+      1: { gap: vars.size.s1 },
+      2: { gap: vars.size.s2 },
+      3: { gap: vars.size.s3 },
+      4: { gap: vars.size.s4 }
+    }
+  }
+});
--- a/packages/components/src/components/common/toggleGroup/toggleGroup.tsx
+++ b/packages/components/src/components/common/toggleGroup/toggleGroup.tsx
@@ -1,0 +1,24 @@
+import * as React from 'react';
+import * as ToggleGroupPrimitive from '@radix-ui/react-toggle-group';
+
+import { cn } from '@/utils/class-name';
+import { toggleGroup } from './toggleGroup.css';
+
+type ToggleGroupProps = React.ComponentPropsWithoutRef<typeof ToggleGroupPrimitive.Root> & {
+  gap?: 1 | 2 | 3 | 4 | undefined;
+};
+
+const ToggleGroup = React.forwardRef<React.ElementRef<typeof ToggleGroupPrimitive.Root>, ToggleGroupProps>(
+  ({ className, gap, ...props }, ref) => (
+    <ToggleGroupPrimitive.Root className={cn(toggleGroup({ gap }), className, 'ui-toggle-group')} {...props} ref={ref} />
+  )
+);
+ToggleGroup.displayName = ToggleGroupPrimitive.Root.displayName;
+
+const ToggleGroupItem = React.forwardRef<
+  React.ElementRef<typeof ToggleGroupPrimitive.Item>,
+  React.ComponentPropsWithoutRef<typeof ToggleGroupPrimitive.Item>
+>(({ className, ...props }, ref) => <ToggleGroupPrimitive.Item ref={ref} className={cn(className, 'ui-toggle-group-item')} {...props} />);
+ToggleGroupItem.displayName = ToggleGroupPrimitive.Item.displayName;
+
+export { ToggleGroup, ToggleGroupItem };
--- a/package-lock.json
+++ b/package-lock.json
@@ -4688,6 +4688,350 @@
         }
       }
     },
+    "node_modules/@radix-ui/react-toggle": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/@radix-ui/react-toggle/-/react-toggle-1.1.0.tgz",
+      "integrity": "sha512-gwoxaKZ0oJ4vIgzsfESBuSgJNdc0rv12VhHgcqN0TEJmmZixXG/2XpsLK8kzNWYcnaoRIEEQc0bEi3dIvdUpjw==",
+      "dependencies": {
+        "@radix-ui/primitive": "1.1.0",
+        "@radix-ui/react-primitive": "2.0.0",
+        "@radix-ui/react-use-controllable-state": "1.1.0"
+      },
+      "peerDependencies": {
+        "@types/react": "*",
+        "@types/react-dom": "*",
+        "react": "^16.8 || ^17.0 || ^18.0 || ^19.0 || ^19.0.0-rc",
+        "react-dom": "^16.8 || ^17.0 || ^18.0 || ^19.0 || ^19.0.0-rc"
+      },
+      "peerDependenciesMeta": {
+        "@types/react": {
+          "optional": true
+        },
+        "@types/react-dom": {
+          "optional": true
+        }
+      }
+    },
+    "node_modules/@radix-ui/react-toggle-group": {
+      "version": "1.1.0",
+      "resolved": "https://registry.npmjs.org/@radix-ui/react-toggle-group/-/react-toggle-group-1.1.0.tgz",
+      "integrity": "sha512-PpTJV68dZU2oqqgq75Uzto5o/XfOVgkrJ9rulVmfTKxWp3HfUjHE6CP/WLRR4AzPX9HWxw7vFow2me85Yu+Naw==",
+      "dependencies": {
+        "@radix-ui/primitive": "1.1.0",
+        "@radix-ui/react-context": "1.1.0",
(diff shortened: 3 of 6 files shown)


ivy-lli (commented):
Nice, some points:
- I would still call it ToggleGroup and not ToggleBar
- Now we have duplicated styles from the button. This and the whole `ToggleBarItemOnlyIcon` is not really needed... We chouls also just use the `asChild` prop from the ToggleGroup.Item. If we do it like that for the beginning, I think we can delete most of the css.
```tsx
export const Default: Story = {
  render: props => (
    <ToggleBar {...props} aria-label='Class selection'>
      <ToggleBarItem value='data-class' asChild>
        <Button>Option 1</Button>
      </ToggleBarItem>
      <ToggleBarItem value='business-data-class' asChild>
        <Button icon={IvyIcons.AlignVertical}>Option 2</Button>
      </ToggleBarItem>
      <ToggleBarItem value='entity-class' asChild>
        <Button>Option 1</Button>
      </ToggleBarItem>
    </ToggleBar>
  )
};

export const OnlyIcon: Story = {
  render: props => (
    <ToggleBar {...props} aria-label='Class selection'>
      <ToggleBarItem value='desktop' asChild>
        <Button icon={IvyIcons.EventStart} size='large' />
      </ToggleBarItem>
      <ToggleBarItem value='tablet' asChild>
        <Button icon={IvyIcons.LaneSwimlanes} size='large' />
      </ToggleBarItem>
      <ToggleBarItem value='mobile' asChild>
        <Button icon={IvyIcons.ChangeType} size='large' />
      </ToggleBarItem>
    </ToggleBar>
  )
};
```

- Only if you don't are aware, most of the components are from shadcn/ui with some adjustments: https://ui.shadcn.com/docs/components/toggle-group

ivy-lli:
Technically there is also a reason why this component didn't exist. Because it is really easy to implement it by your own just with the toggle function of a button, but I think it is fine to have it :)

ivy-edp:
@ivy-lli Thanks for the feedback, i will try to implement it 👍 
Regarding the renaming: I initially wanted to name the component ToggleGroup, however I am unable to name it like that, as the Component from radix-ui is called "ToggleGroup" already. I dont know why because all the other components are with ...Primitive at the and:
grafik

Yes, I also thought that it might be a bit overkill to create this ToggleBar component for it, but it was also a good exercise for me to familiarise myself with the UI components :)

packages/components/src/components/common/toggleGroup/toggleGroup.tsx
```diff
@@ -0,0 +1,41 @@
+import * as React from 'react';
+import * as ToggleGroupPrimitive from '@radix-ui/react-toggle-group';
+
+import { cn } from '@/utils/class-name';
+import { toggleBar } from './toggleGroup.css';
+import { useReadonly } from '@/context';
+import { useField } from '@/components';
+
+type ToggleBarProps = React.ComponentPropsWithoutRef<typeof ToggleGroupPrimitive.Root> & {
```
ivy-lli, line 9: rename

packages/components/src/components/common/toggleGroup/toggleGroup.tsx
```diff
@@ -0,0 +1,41 @@
...
+  gap?: 1 | 2 | 3 | 4 | undefined;
+};
+
+const ToggleGroup = React.forwardRef<React.ElementRef<typeof ToggleGroupPrimitive.Root>, ToggleBarProps>(
+  ({ disabled, className, gap, ...props }, ref) => {
+    const readonly = useReadonly();
+    return (
+      <ToggleGroupPrimitive.Root
+        className={cn(toggleBar({ gap }), className, 'ui-toggle-bar')}
```
ivy-lli, line 18: rename

packages/components/src/components/common/toggleGroup/toggleGroup.tsx
```diff
@@ -0,0 +1,41 @@
...
+);
+ToggleGroup.displayName = ToggleGroupPrimitive.Root.displayName;
+
+const ToggleGroupItem = React.forwardRef<
+  React.ElementRef<typeof ToggleGroupPrimitive.Item>,
+  React.ComponentPropsWithoutRef<typeof ToggleGroupPrimitive.Item>
+>(({ className, ...props }, ref) => {
+  const { inputProps } = useField();
+  return (
+    <ToggleGroupPrimitive.Item ref={ref} className={cn(className, 'ui-toggle-bar-item')} {...inputProps} {...props}>
```
ivy-lli, line 34: rename

packages/components/src/components/common/toggleGroup/toggleGroup.tsx
```diff
@@ -0,0 +1,41 @@
...
+    );
+  }
+);
+ToggleGroup.displayName = ToggleGroupPrimitive.Root.displayName;
+
+const ToggleGroupItem = React.forwardRef<
+  React.ElementRef<typeof ToggleGroupPrimitive.Item>,
+  React.ComponentPropsWithoutRef<typeof ToggleGroupPrimitive.Item>
+>(({ className, ...props }, ref) => {
+  const { inputProps } = useField();
```
ivy-lli, line 32: I think we don't use this component inside a field. so this is not needed

packages/components/src/components/common/toggleGroup/toggleGroup.css.ts
```diff
@@ -0,0 +1,21 @@
+import { vars } from '@/styles/theme.css';
+import { recipe } from '@vanilla-extract/recipes';
+
+export const toggleBar = recipe({
```
ivy-lli, line 4: rename

packages/components/src/components/common/toggleGroup/toggleGroup.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { useReadonly } from '@/context';
+import { useField } from '@/components';
+
+type ToggleBarProps = React.ComponentPropsWithoutRef<typeof ToggleGroupPrimitive.Root> & {
+  gap?: 1 | 2 | 3 | 4 | undefined;
+};
+
+const ToggleGroup = React.forwardRef<React.ElementRef<typeof ToggleGroupPrimitive.Root>, ToggleBarProps>(
+  ({ disabled, className, gap, ...props }, ref) => {
+    const readonly = useReadonly();
```
ivy-lli, line 15: not sure if this component need to listen to the global readonly context...
Does this need to be disabled, if e.g. the form editor is readonly? I think not, so I would remove it

ivy-lli (commented):
Don't forget to export it in the components/index.ts file

packages/components/src/components/common/toggleGroup/toggleGroup.tsx
ivy-lli: yes it can. just set the disabled flag ;)

ivy-edp:
@ivy-lli (Maybe you know) Did I forget something in or regarding the components/index.ts file or why does the pipeline suddenly fail?

packages/components/src/components/common/toggleGroup/toggleGroup.tsx
```diff
@@ -0,0 +1,37 @@
...
+      <ToggleGroupPrimitive.Root
+        className={cn(toggleGroup({ gap }), className, 'ui-toggle-group')}
+        disabled={disabled}
+        {...props}
+        ref={ref}
+      />
+    );
+  }
+);
+ToggleGroup.displayName = ToggleGroupPrimitive.Root.displayName;
```
ivy-lli, line 23: as you no longer do something with the disabled prop, you can just remove it. Then it will be part of the ...props.
You also no longer do something else than return jsx so you can use the shorthand syntax:

```tsx
const ToggleGroup = React.forwardRef<React.ElementRef<typeof ToggleGroupPrimitive.Root>, ToggleGroupProps>(
  ({ className, gap, ...props }, ref) => (
    <ToggleGroupPrimitive.Root
      className={cn(toggleGroup({ gap }), className, 'ui-toggle-group')}
      {...props}
      ref={ref}
    />
  )
);
ToggleGroup.displayName = ToggleGroupPrimitive.Root.displayName;
```

packages/components/src/components/common/toggleGroup/toggleGroup.tsx
```diff
@@ -0,0 +1,37 @@
...
+  React.ElementRef<typeof ToggleGroupPrimitive.Item>,
+  React.ComponentPropsWithoutRef<typeof ToggleGroupPrimitive.Item>
+>(({ className, ...props }, ref) => {
+  return (
+    <ToggleGroupPrimitive.Item ref={ref} className={cn(className, 'ui-toggle-group-item')} {...props}>
+      {props.children}
+    </ToggleGroupPrimitive.Item>
+  );
+});
+ToggleGroupItem.displayName = ToggleGroupPrimitive.Item.displayName;
```
ivy-lli, line 35: the children is already part of props, so just don't do something with it ;)
and same here directly return and save one indent:
```tsx
const ToggleGroupItem = React.forwardRef<
  React.ElementRef<typeof ToggleGroupPrimitive.Item>,
  React.ComponentPropsWithoutRef<typeof ToggleGroupPrimitive.Item>
>(({ className, ...props }, ref) => <ToggleGroupPrimitive.Item ref={ref} className={cn(className, 'ui-toggle-group-item')} {...props} />);
ToggleGroupItem.displayName = ToggleGroupPrimitive.Item.displayName;
```

ivy-lli (approved):
Nice. Much less code right? 😉

ivy-edp:
Yes, thanks for the great improvement tips 👏

PR #114, opened 2024-07-03
XIVY-13261 Remove Padding from toggleGroup

ivy-edp:
I realised during the integration into the form editor that I have to remove this padding, sorry

### DIFF
--- a/packages/components/src/components/common/toggleGroup/toggleGroup.css.ts
+++ b/packages/components/src/components/common/toggleGroup/toggleGroup.css.ts
@@ -6,7 +6,6 @@ export const toggleGroup = recipe({
     display: 'inline-flex',
     alignItems: 'center',
     gap: vars.size.s2,
-    padding: '10px',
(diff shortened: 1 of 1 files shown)


packages/components/src/components/common/toggleGroup/toggleGroup.css.ts
```diff
@@ -6,7 +6,6 @@ export const toggleGroup = recipe({
     display: 'inline-flex',
     alignItems: 'center',
     gap: vars.size.s2,
-    padding: '10px',
     backgroundColor: vars.color.background,
     borderRadius: vars.border.r2
```
ivy-lli, line 10: maybe also useless?

ivy-edp: you are right, I will remove these in another PR. However, I no longer sit there as a reviewer because the change is too small.

PR #118, opened 2024-07-08
XIVY-13261 Add minWidth to ToolbarContainer

ivy-edp:
I have extended the ToolbarContainer component so that I can easily enable a similar responsive behaviour of the toolbar analogous to the process editor. Now it is not only possible to define that parts above a certain width are no longer displayed, but also the opposite, i.e. that parts above a certain width are only displayed.


ivy-edp:
With this I can adjust the Components-Part in the Toolbar like this and achive a even greater responsiveness:
grafik

ivy-lli (approved):

PR #124, opened 2024-07-16
Fix toolbar container queries

ivy-lli:
@ivy-edp 
Always make sure that you catch all cases otherwise something like the following can happen:
image


ivy-edp:
Oh sorry, thanks 👍

PR #125, opened 2024-07-17
Improve browser component

ivy-lli:
- Support for load more rows lazy
- Add header property for more info or a filter or whatever (imporve story with info what which browser do special)
- Export more Browser types
- Change Browser type `BrowserValue` to `BrowserResult` (with `value` attribute instead of `cursor`)
- Change apply function signature from `(value?: BrowserValue, type?: string) => void` to `(browserName: string, result?: BrowserResult) => void`
- Change applyModifier signature from `(value: string) => BrowserValue` to `(row: Row<BrowserNode>) => BrowserResult`


weissreto (approved):
As far as I understand the code it looks good to me.

PR #138, opened 2024-08-20
Feedback of anais

ivy-lli:
- Dialog (increase gap, padding, border radius, close button size)
- Button (border radius)
- Select (trigger icon size)


ivy-cst (commented):
That was fast :+1:

PR #144, opened 2024-08-26
XIVY-12288 add config editor

ivy-lgi:


ivy-lgi:
image

ivy-lli (commented):
There is a reason why I didn't make a component for this but an example story...
IMO every editor should follow the same shape yes but should also have the flexibility to change every part, so I think provide each part to quickly generate this layout is enough and we should not force an editor to have this "corset".

ivy-lli:
So I'm definitly against this component, at least at the moment.
If we see at some point that e.g. the rest / webservice / variable editor all share the same layout then we can talk about this again. But I think all this editors will live in the config editor repo so this component should then live there.

ivy-lgi:
> So I'm definitly against this component, at least at the moment. If we see at some point that e.g. the rest / webservice / variable editor all share the same layout then we can talk about this again. But I think all this editors will live in the config editor repo so this component should then live there.

I see. I was thinking that all the editors will now go into their own repo but I guess only the Data Class Editor is separate from all the Config Editors?
As far as I can tell, the Config Editors will all share the same layout, apart from the titles, the detail toolbar icon, and possibly what's present in the settings menu. I thought of making it possible to provide these as arguments to the component. But as you said, if we do this, we will do it in the Config Editor repo.

PR #145, opened 2024-08-27
XIVY-12288 add table utils

ivy-lgi:
Copied over from config-editor-client

### DIFF
--- a/packages/components/src/utils/table/table-data.test.ts
+++ b/packages/components/src/utils/table/table-data.test.ts
@@ -1,0 +1,38 @@
+import { beforeEach, describe, expect, test } from 'vitest';
+import { updateRowData } from './table-data';
+import type { TestData } from './test-utils/types';
+
+let data: Array<TestData>;
+let newRowData: TestData;
+
+beforeEach(() => {
+  data = [
+    { name: 'NameData0', value: 'ValueData0' },
+    { name: 'NameData1', value: 'ValueData1' },
+    { name: 'NameData2', value: 'ValueData2' }
+  ];
+  newRowData = { name: 'newDataName', value: 'newDataValue' };
+});
+
+describe('table-data', () => {
+  describe('updateRowData', () => {
+    test('default', () => {
+      const originalData = structuredClone(data);
+      const newData = updateRowData(data, 1, newRowData);
+      expect(data).toEqual(originalData);
+      expect(newData).not.toBe(data);
+      expect(newData).toHaveLength(3);
+      expect(newData[0]).toEqual(data[0]);
+      expect(newData[1]).toEqual(newRowData);
+      expect(newData[2]).toEqual(data[2]);
+    });
+
+    test('indexOutOfBounds', () => {
+      const originalData = structuredClone(data);
+      const newData = updateRowData(data, 3, newRowData);
+      expect(data).toEqual(originalData);
+      expect(newData).not.toBe(data);
+      expect(newData).toEqual(data);
+    });
+  });
+});
--- a/packages/components/src/utils/table/table-data.test.ts
+++ b/packages/components/src/utils/table/table-data.test.ts
@@ -1,38 +1,28 @@
-import { beforeEach, describe, expect, test } from 'vitest';
+import { setupData } from '@/utils/table/test-utils/setup';
+import { describe, expect, test } from 'vitest';
 import { updateRowData } from './table-data';
-import type { TestData } from './test-utils/types';
 
-let data: Array<TestData>;
-let newRowData: TestData;
+const newRowData = { name: 'newDataName', value: 'newDataValue' };
 
-beforeEach(() => {
-  data = [
-    { name: 'NameData0', value: 'ValueData0' },
-    { name: 'NameData1', value: 'ValueData1' },
-    { name: 'NameData2', value: 'ValueData2' }
-  ];
-  newRowData = { name: 'newDataName', value: 'newDataValue' };
-});
+describe('updateRowData', () => {
+  test('default', () => {
+    const data = setupData();
+    const originalData = structuredClone(data);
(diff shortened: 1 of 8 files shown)


packages/components/src/utils/table/table.ts
```diff
@@ -0,0 +1,49 @@
+import type { Row, Table } from '@tanstack/react-table';
+
+export const selectRow = <TData>(table: Table<TData>, rowId?: string) => {
+  if (!rowId || rowId === '') {
+    table.setRowSelection({});
+  } else {
+    table.setRowSelection({ [`${rowId}`]: true });
```
ivy-lli, line 7: why not simply use { [rowId]: true }?

ivy-lli: I would make a singleliner out of this, and because of this I'm not sure about the real utilize of this function. Is it really needed to have a simplification of something which is already simple?

`table.setRowSelection(rowId === undefined || row === '' ? {} : {[rowId]: true})`

packages/components/src/utils/table/table.ts
```diff
@@ -0,0 +1,49 @@
...
+  }
+};
+
+export const getFirstSelectedRow = <TData>(table: Table<TData>) => {
+  return getRow(table, Object.keys(table.getState().rowSelection)[0]);
+};
```
ivy-lli, line 13: Not sure why we need to create our own function if tanstack table has something for this out of the box?
`table.getSelectedRowModel().flatRows[0]`

packages/components/src/utils/table/table.ts
```diff
@@ -0,0 +1,49 @@
...
+const adjustSelectionAfterDeletionOfRow = <TData>(data: Array<TData>, table: Table<TData>, row: Row<TData>) => {
+  if (data.length === 0) {
+    selectRow(table);
+  } else if (row.index === data.length) {
+    selectRow(table, String(data.length - 1));
+  }
+};
+
+/* workaround for "table.getIsSomeRowsSelected" as it returns false if only last remaining row is selected */
+export const isRowSelected = <TData>(table: Table<TData>) => Object.keys(table.getState().rowSelection).length > 0;
```
ivy-lli, line 49: also why not simply use `table.getSelectedRowModel().rows.lenght > 0`

ivy-lli: not sure if we should create a function for something just simple as this use case

packages/components/src/utils/table/table.test.ts
```diff
@@ -0,0 +1,136 @@
...
+let table: Table<TestData>;
+let onRowSelectionChangeValue: Updater<RowSelectionState>;
+
+beforeEach(() => {
+  data = [
+    { name: 'NameData0', value: 'ValueData0' },
+    { name: 'NameData1', value: 'ValueData1' },
+    { name: 'NameData2', value: 'ValueData2' }
+  ];
+  newRowData = { name: 'newDataName', value: 'newDataValue' };
```
ivy-lli, line 19: same here as above. no need to have this in a before each I think...

packages/components/src/utils/table/table.test.ts
```diff
@@ -0,0 +1,136 @@
...
+    data: data,
+    getCoreRowModel: getCoreRowModel(),
+    onStateChange: () => {},
+    onRowSelectionChange: (value: Updater<RowSelectionState>) => {
+      onRowSelectionChangeValue = value;
+    },
+    renderFallbackValue: undefined,
+    state: {}
+  });
+  onRowSelectionChangeValue = {};
```
ivy-lli, line 32: also instead of hold this in a global let, better create a setup function which you call at the start of each test.

const { table, onRowSelectionChangeValue } = setupTable()

packages/components/src/utils/table/test-utils/types.ts
```diff
@@ -0,0 +1,4 @@
+export interface TestData {
+  name: string;
+  value: string;
+}
```
ivy-lli, line 4: if you don't use the beforeeach stuff I think you don't need this type at all...

ivy-lli (changes requested):

packages/components/src/utils/table/table.ts
ivy-lgi: I like the current implementation better because it is much easier to read.
I also like that this is a defined function, again because of readability when this is needed. This function is currently called 4 different times in the Variables Editor.

packages/components/src/utils/table/table.test.ts
ivy-lgi: Same answer as above ^^

packages/components/src/utils/table/table-data.test.ts
```diff
@@ -0,0 +1,38 @@
...
+let newRowData: TestData;
+
+beforeEach(() => {
+  data = [
+    { name: 'NameData0', value: 'ValueData0' },
+    { name: 'NameData1', value: 'ValueData1' },
+    { name: 'NameData2', value: 'ValueData2' }
+  ];
+  newRowData = { name: 'newDataName', value: 'newDataValue' };
+});
```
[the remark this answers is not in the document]
ivy-lli: A followup test should never fail if a previous test fails!
The problem with this let and beforeEach is that let is in this case global for each test, but what if the tests run in parallel and some test do change this data? then an other test fail because of this, which is exactly the reason for flaky and hard to analyze test failures...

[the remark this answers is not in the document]
ivy-lgi: I see, makes sense. I guess in this case I am adding this data array to the return value of `setupTable` as well. This way, a new array is created for every test. Is this fine with you?

ivy-lli (approved):
Yes, thanks for the adjustments 🥇 
Sorry to be sometimes a little picky 🙈

PR #151, opened 2024-09-06
XIVY-12289 return selection after deletion

ivy-lgi:
When calling `deleteFirstSelectedRow` the selection in the table is changed by using some logic. The value of the new selection is usually needed after calling this function. E.g. in the data class editor, after deleting the selected row, the selected field needs to be set to the same selection in the table. Instead of recalculating this value using the same logic, it will be returned from `deleteFirstSelectedRow`.

### DIFF
--- a/packages/components/src/utils/table/table.test.ts
+++ b/packages/components/src/utils/table/table.test.ts
@@ -42,12 +42,13 @@ describe('deleteFirstSelectedRow', () => {
     const { data, table, onRowSelectionChangeValues } = setupTable();
     const originalData = structuredClone(data);
     table.getState().rowSelection = { '1': true };
-    const newData = deleteFirstSelectedRow(table, data);
+    const { newData, selection } = deleteFirstSelectedRow(table, data);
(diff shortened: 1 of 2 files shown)


ivy-lgi:
```typescript
  const deleteField = () => {
    const newFields = deleteFirstSelectedRow(table, dataClass.fields);
    const newDataClass = structuredClone(dataClass);
    newDataClass.fields = newFields;
    setDataClass(newDataClass);
    if (newFields.length === 0) {
      setSelectedField();
    } else if (selectedField === newFields.length) {
      setSelectedField(selectedField - 1);
    }
  };
```
turns into:
```typescript
  const deleteField = () => {
    const { newData: newFields, selection } = deleteFirstSelectedRow(table, dataClass.fields);
    const newDataClass = structuredClone(dataClass);
    newDataClass.fields = newFields;
    setDataClass(newDataClass);
    setSelectedField(selection);
  };
```

ivy-lli (approved):

PR #155, opened 2024-09-11
XIVY-13131 Make row not selectable & add custom expandedState

ivy-edp:
For the Logic Browser, I have implemented the following two functions:
- Control which row is selectable and which is not
- Define a custom expandedState

This is because the Logic Browser always has the same two root nodes, "Methods" and "Events," and these should always be expanded by default and not selectable.


ivy-lli (commented):
👍

PR #161, opened 2024-09-19
XIVY-12288 button props in button group

ivy-lgi:
I want to use the `disabled` property of the buttons in the control to disable this delete button when no annotation is selected:

image

Any reason to not allow using all the button properties?

I guess otherwise I could hide it entirely or maybe solve this with CSS.


ivy-lli (approved):
Nope. Just never used the ButtonGroup at all. 
Honestly I'm not sure, why it exists anyways... IMO not really worth to have this component as the implementation is nearly nothing :D

PR #163, opened 2024-09-20
Add BasicCollapsible

ivy-lli:
Basic select that supports to have message, controls and change it's open state if the `open` prop changes or has default open if there are messages


ivy-lgi (commented):
nice :+1:

PR #169, opened 2024-10-09
XIVY-15138 add useShortcut hook

ivy-lgi:
I think this is everything needed here since I can solve the `Enter` case with a form.

I'll answer the question from the [other PR](https://github.com/axonivy/dataclass-editor-client/pull/48) about the following code :
```
  const callbackRef = useRef(callback);
  useLayoutEffect(() => {
    callbackRef.current = callback;
  });
```
I got this from [this article](https://www.epicreact.dev/the-latest-ref-pattern-in-react) and in short terms it allows a consumer of this hook to not care about setting up the `callback` function in a `useCallback`. Because we use a reference we do not need to add it to the dependency array of `handleKeyPress`.

### DIFF
--- a/packages/components/src/utils/useShortcut.test.tsx
+++ b/packages/components/src/utils/useShortcut.test.tsx
@@ -1,0 +1,62 @@
+import { useShortcut } from '@/utils/useShortcut';
+import { fireEvent, render } from '@testing-library/react';
+
+type TestComponentProps = {
+  shortcutKey: string;
+  callback: () => void;
+  modifiers?: { ctrl?: boolean; alt?: boolean; shift?: boolean };
+};
+
+const TestComponent = ({ shortcutKey: key, callback, modifiers }: TestComponentProps) => {
+  useShortcut(key, callback, modifiers);
+  return <div>Test Component</div>;
+};
+
+test('default', () => {
+  let shortcutTriggered = false;
+  render(<TestComponent shortcutKey={'a'} callback={() => (shortcutTriggered = true)} />);
+
+  fireEvent.keyDown(document, { key: 'a', ctrlKey: true, shiftKey: true });
+  expect(shortcutTriggered).toBeFalsy();
+
+  fireEvent.keyDown(document, { key: 'a', altKey: true, shiftKey: true });
+  expect(shortcutTriggered).toBeFalsy();
+
+  fireEvent.keyDown(document, { key: 'z', ctrlKey: true, altKey: true });
+  expect(shortcutTriggered).toBeFalsy();
+
+  fireEvent.keyDown(document, { key: 'a', ctrlKey: true, altKey: true });
+  expect(shortcutTriggered).toBeTruthy();
+});
+
+test('custom', () => {
+  let shortcutTriggered = false;
+  render(<TestComponent shortcutKey={'a'} callback={() => (shortcutTriggered = true)} modifiers={{ shift: true }} />);
+
+  fireEvent.keyDown(document, { key: 'a', ctrlKey: true });
+  expect(shortcutTriggered).toBeFalsy();
+
+  fireEvent.keyDown(document, { key: 'a', altKey: true });
+  expect(shortcutTriggered).toBeFalsy();
+
+  fireEvent.keyDown(document, { key: 'z', shiftKey: true });
+  expect(shortcutTriggered).toBeFalsy();
+
+  fireEvent.keyDown(document, { key: 'a', shiftKey: true });
+  expect(shortcutTriggered).toBeTruthy();
+});
+
+test('mac', () => {
+  Object.defineProperty(window.navigator, 'userAgent', {
+    value: 'Mac'
+  });
+
+  let shortcutTriggered = false;
+  render(<TestComponent shortcutKey={'a'} callback={() => (shortcutTriggered = true)} />);
+
+  fireEvent.keyDown(document, { key: 'a', ctrlKey: true, altKey: true });
+  expect(shortcutTriggered).toBeFalsy();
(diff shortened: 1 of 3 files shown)


packages/components/src/utils/useShortcut.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export const useShortcut = (key: string, callback: () => void) => {
+  const callbackRef = useRef(callback);
+  useLayoutEffect(() => {
+    callbackRef.current = callback;
+  });
+
+  const handleKeyPress = useCallback(
+    (event: KeyboardEvent) => {
+      if (event.ctrlKey && event.altKey && event.key === key) {
```
ivy-lli, line 11: what is when we want to add a shortcut with other modifiers?

ivy-lli: And as also said on the other PR, mac uses mostly cmd for short cuts and not crtl. Correct me if I'm wrong @ivy-lmu

ivy-lgi: > what is when we want to add a shortcut with other modifiers?

I think we should not add a shortcut with other modifiers if at all possible. For consistency in the way our custom shortcuts are accessed and because this is the least used modifier combination by browsers. However, I am fine with adding an optional `options` argument to change the modifiers but keeping this as the default just in case.

ivy-lli: Not sure about this. E.g. Yesterday we talked about the idea if you are in the new dialog and e.g. press enter with the ctrl key (or whatever) then a new thing gets created without leaving the dialog, so you can directly add an additional one.

ivy-lli: I'm not saing that the default should not be with ctrl + alt.

ivy-lli: e.g. glsp does something like the following:

```ts
export function matchesKeystroke(event: KeyboardEvent, code: KeyCode, ...modifiers: KeyboardModifier[]): boolean {
    if (getActualCode(event) !== code)
        return false;
    if (isMac()) {
        if (event.ctrlKey !== (modifiers.findIndex(m => m === 'ctrl') >= 0))
            return false;
        if (event.metaKey !== (modifiers.findIndex(m => m === 'meta' || m === 'ctrlCmd') >= 0))
            return false;
    } else {
        if (event.ctrlKey !== (modifiers.findIndex(m => m === 'ctrl' || m === 'ctrlCmd') >= 0))
            return false;
        if (event.metaKey !== (modifiers.findIndex(m => m === 'meta') >= 0))
            return false;
    }
    if (event.altKey !== (modifiers.findIndex(m => m === 'alt') >= 0))
        return false;
    if (event.shiftKey !== (modifiers.findIndex(m => m === 'shift') >= 0))
        return false;
    return true;
}

export function isMac(): boolean {
    return window.navigator.userAgent.indexOf("Mac") !== -1;
}
```

at least as hint how to do this cmd or ctrl button ;)

packages/components/src/utils/useShortcut.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const useShortcut = (key: string, callback: () => void, modifiers?: { ctrl?: boolean; alt?: boolean; shift?: boolean }) => {
+  const callbackRef = useRef(callback);
+  useLayoutEffect(() => {
+    callbackRef.current = callback;
+  });
+
+  const handleKeyPress = useCallback(
+    (event: KeyboardEvent) => {
+      const ctrl = modifiers ? modifiers.ctrl : true;
```
ivy-lli, line 11: `const ctrl = modifiers?.ctrl ?? true;`

packages/components/src/utils/useShortcut.test.tsx
```diff
@@ -0,0 +1,62 @@
...
+const TestComponent = ({ shortcutKey: key, callback, modifiers }: TestComponentProps) => {
+  useShortcut(key, callback, modifiers);
+  return <div>Test Component</div>;
+};
+
+test('default', () => {
+  let shortcutTriggered = false;
+  render(<TestComponent shortcutKey={'a'} callback={() => (shortcutTriggered = true)} />);
+
+  fireEvent.keyDown(document, { key: 'a', ctrlKey: true, shiftKey: true });
```
ivy-lli, line 19: Isn't fireEvent deprecated and userEvent should be used?

ivy-lli (approved):
https://testing-library.com/docs/dom-testing-library/api-events/

packages/components/src/utils/useShortcut.ts
ivy-lgi: It's not the same thing in this case.
If no `modifiers` object is provided I want the defaults to be used (`true`, `true`, `false`).
If a `modifiers` object is provided, I want to use the values it contains or `false` for missing values.

With the current implementation `modifiers = {shift: true}` results in: `{ctrl: false, alt: false, shift: true}`
With your change it would result in: `{ctrl: true, alt: true, shift:true}` because it's going to use the defaults for the missing values.

Of course we could change the behaviour but I think this makes more sense or make the properties of modifiers all required in the first place.

packages/components/src/utils/useShortcut.test.tsx
ivy-lgi: Not deprecated it seems but I just found this remark [in the doc](https://testing-library.com/docs/dom-testing-library/api-events/):
> Most projects have a few use cases for fireEvent, but the majority of the time you should probably use [@testing-library/user-event](https://testing-library.com/docs/user-event/intro).

I'll check.

PR #171, opened 2024-10-10
Fix dnd line if move down

ivy-lli:
Screen Recording 2024-10-10 at 15 03 20


ivy-lgi (approved):
Very nice, thanks :+1:

PR #175, opened 2024-10-18
XIVY-14728 implements component for inputTag

ivy-fhe:
Implementation of a generic version of **CardText** from inscription-client
Accepts following props:
```typescript
export type InputBadgeProps = {
  value: string;
  badgeProps: Array<BadgeProps>;
};

type BadgeProps = { regex: string; icon: IvyIcons; badgeTextGen: (text: string) => string };
```
image


packages/components/src/components/common/inputtag/inputTag.tsx
```diff
@@ -0,0 +1,61 @@
+import { forwardRef, type ElementRef } from 'react';
+import { useField } from '../field/field';
+import { IvyIcon } from '../icon/icon';
+import './inputTag.css';
+
+import { type IvyIcons } from '@axonivy/ui-icons';
+
+export type InputTagProps = {
+  tags?: string[];
+  lines?: string[][];
```
ivy-lli, line 10: this will not be the input I think.
The input is just a string, single line for inputs, multiline for areas.

And the tagOrText function (which I would call `findTags` or something like that), should find all tags in it

ivy-lli: E.g. the text `This is a <%= ivy.log.info() %> text with a tag` should then result in `This is a [ivy tag] text`.
Of course also with multiple tags...

packages/components/src/components/common/inputtag/inputTag.css
ivy-lli, line 1: we work with vanilla css in this repo, so we have typesafe and unique style rules

ivy-lli (changes requested):
I'm not sure if tag is the right name... 
I think this are badges. Because we have already tags in the inscription in the general section

packages/components/src/components/common/inputtag/inputBadge.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+const badgeRegex = [
+  {
+    delimiter: {
+      start: '<%=',
+      end: '%>'
+    },
+    badgeTextGen: (text: string) => {
+      return text;
+    },
```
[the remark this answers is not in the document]
ivy-lli: maybe it would be also nice if one of this badgeTextGen would actually do something :)

packages/components/src/components/common/inputtag/inputBadge.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { inputBadge, inputBadgeIcon, inputBadgeLine, inputBadgeOutput } from './inputBadge.css';
+
+import type { IvyIcons } from '@axonivy/ui-icons';
+
+export type InputBadgeProps = {
+  value: string;
+  badgeRegex: BadgeRegex[];
+};
+
+type BadgeRegex = { delimiter: { start: string; end: string }; icon: IvyIcons; badgeTextGen: (text: string) => string };
```
ivy-lli, line 13: The term regex is for me reserved, and what you do here is not really a "regex" type.
maybe we should call this type something like `BadgeExpression` or `BadgeBlock`

packages/components/src/components/common/inputtag/inputBadge.tsx
```diff
@@ -0,0 +1,79 @@
+import { forwardRef, type ElementRef } from 'react';
+import { useField } from '../field/field';
+import { IvyIcon } from '../icon/icon';
+import { inputBadge, inputBadgeIcon, inputBadgeLine, inputBadgeOutput } from './inputBadge.css';
+
+import type { IvyIcons } from '@axonivy/ui-icons';
+
+export type InputBadgeProps = {
+  value: string;
+  badgeRegex: BadgeRegex[];
```
ivy-lli, line 10: I prefer the `Array<BadgeRegex>` syntax. As it more reads like `an array of type` instead of `type as array`

packages/components/src/components/common/inputtag/inputBadge.tsx
```diff
@@ -0,0 +1,79 @@
...
+};
+
+type BadgeRegex = { delimiter: { start: string; end: string }; icon: IvyIcons; badgeTextGen: (text: string) => string };
+
+export type BadgeTextProps = { icon: IvyIcons | undefined; text: string };
```
ivy-lli, line 15: if you want an optional parameter it is better to use the `?` operator. 
Because if you do it with `| undefined` you have to pass an argument, at least `undefined`. e.g `{icon: undefined, text: 'hi'}`
But if you do it with the `?`, `{icon?: IvyIcons, text: string}`, then you don't have to pass anything for this attribute: `{text: 'hi'}`

packages/components/src/components/common/inputtag/inputBadge.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+const Badge = ({ badge }: { badge: BadgeTextProps }) => {
```
ivy-lli, line 17: why wrap it again in a badge object and not directly use the properties?
```suggestion
const Badge = ({ text, icon }: BadgeTextProps ) => {
```

packages/components/src/components/common/inputtag/inputBadge.css.ts
```diff
@@ -0,0 +1,39 @@
...
+  color: vars.color.body,
+  padding: '2px 0.24em',
+  backgroundColor: vars.color.n75,
+  borderRadius: vars.border.r1,
+  userSelect: 'none',
+  border: vars.border.basic
+});
+
+export const inputBadgeIcon = style({
+  marginRight: vars.size.s1,
```
ivy-lli, line 34: use flex with gap for things like that.

packages/components/src/components/common/inputtag/inputBadge.tsx
```diff
@@ -0,0 +1,79 @@
...
+    return (
+      <span className={inputBadge}>
+        <span className={inputBadgeIcon}>
+          <IvyIcon icon={badge.icon} />
+        </span>
+        {badge.text}
+      </span>
+    );
+  }
+  return <span>{badge.text}</span>;
```
ivy-lli, line 28: I think a text without an icon should not be a badge. So the badge component should only be used for badges.
And I'm also not sure why you need to wrap the icon inside a span...
```suggestion
  (   
    <Flex gap={1} alignItems='center' className={inputBadge}>
      {icon && <IvyIcon icon={badge.icon} />}
      {text}
    </Flex>
  );
```

packages/components/src/components/common/inputtag/inputBadge.tsx
```diff
@@ -0,0 +1,79 @@
...
+      {badges.map((badge, index) => (
+        <Badge key={index} badge={badge} />
+      ))}
+    </output>
+  );
+};
+
+export const InputBadgeArea = forwardRef<ElementRef<'output'>, InputBadgeProps>(({ value, badgeRegex, ...props }, forwardRef) => {
+  const { inputProps } = useField();
+  const lines = value.split(/\r\n|\r|\n/).map(e => findBadges(e, badgeRegex));
```
ivy-lli, line 45: there is already a utility function for this `splitNewLine`

packages/components/src/components/common/inputtag/inputBadge.tsx
```diff
@@ -0,0 +1,79 @@
...
+    </output>
+  );
+});
+
+const findBadges = (value: string, badgeRegex: BadgeRegex[]): BadgeTextProps[] => {
+  const regexStrings: string[] = [];
+  badgeRegex.forEach(reg => {
+    regexStrings.push('(?:' + reg.delimiter.start + '.*?' + reg.delimiter.end + ')');
+  });
+  const separated = value.split(new RegExp('(' + regexStrings.join('|') + ')'));
```
ivy-lli, line 64: we could do this like that, but I can imagine that this is not ideal if we do this everytime. Why shouldn't we just require a regex in the BadgeRegex type? User of this component are devs, I think they should be able to create such a regex, especially if they have examples.

packages/components/src/components/common/inputtag/inputBadge.tsx
```diff
@@ -0,0 +1,79 @@
...
+        {badge.text}
+      </span>
+    );
+  }
+  return <span>{badge.text}</span>;
+};
+
+export const InputBadge = ({ value, badgeRegex, ...props }: InputBadgeProps) => {
+  const { inputProps } = useField();
+  const badges = findBadges(value, badgeRegex);
```
ivy-lli, line 33: this is a more heavy operation I think, so it should be memorized with `useMemo`

packages/components/src/components/common/inputtag/inputBadge.tsx
ivy-lli: should also be memoized

packages/components/src/components/common/inputtag/inputBadge.tsx
```diff
@@ -0,0 +1,79 @@
...
+  let badgeTextProps: BadgeTextProps[] = [];
+  badgeTextProps = separated.map(e => {
+    let prop: BadgeTextProps = { icon: undefined, text: e };
+    badgeRegex.forEach(reg => {
+      if (e && e.startsWith(reg.delimiter.start) && e.endsWith(reg.delimiter.end)) {
+        prop = {
+          icon: reg.icon,
+          text: reg.badgeTextGen(e.replace(reg.delimiter.start, '').replace(reg.delimiter.end, ''))
+        };
+      }
```
ivy-lli, line 74: instead of return a BadgeTextProps array why not directly return an Array of ReactNodes? span's for texts and Badge's for the badges

ivy-lli (changes requested):

packages/components/src/components/common/inputtag/inputBadge.css.ts
```diff
@@ -0,0 +1,39 @@
...
+});
+
+export const inputBadgeLine = style({
+  height: '20px',
+  whiteSpace: 'pre-wrap'
+});
+
+export const inputBadge = style({
+  color: vars.color.body,
+  padding: '2px 0.24em',
```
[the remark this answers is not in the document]
ivy-fhe: yes, I've taken all the styles from the inscription view

packages/components/src/components/common/inputtag/inputBadge.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+  render: props => (
+    <Flex>
+      <InputBadge {...props} />
+    </Flex>
+  )
+};
+
+export const BadgeArea: Story = {
+  render: props => (
+    <Flex>
```
ivy-lli, line 50: this flex needed?

packages/components/src/components/common/inputtag/inputBadge.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const { inputProps } = useField();
+  const items = useMemo(() => findBadges(value, badgeProps), [value, badgeProps]);
+  return (
+    <output className={inputBadgeOutput} {...inputProps} {...props}>
+      {items}
+    </output>
+  );
+};
+
+export const InputBadgeArea = forwardRef<ElementRef<'output'>, InputBadgeProps>(({ value, badgeProps, ...props }, forwardRef) => {
```
ivy-lli, line 34: why is this forwardRef but the normal input not?

packages/components/src/components/common/inputtag/inputBadge.stories.tsx
```diff
@@ -0,0 +1,56 @@
+import type { Meta, StoryObj } from '@storybook/react';
+import { InputBadge, InputBadgeArea } from './inputBadge';
+import { IvyIcons } from '@axonivy/ui-icons';
+import { Flex } from '../flex/flex';
+
+const badgeProps = [
+  {
+    regex: '#{\\s+data\\.[^}]+}',
```
ivy-lli, line 8: Isn't it possible to use a regex type here? 
`regex: /#{...}/`

packages/components/src/components/common/inputtag/inputBadge.tsx
```diff
@@ -0,0 +1,64 @@
...
+      </span>
+    );
+    badgeProps.forEach(prop => {
+      if (text.match(new RegExp(prop.regex))) {
+        node = <Badge key={index} text={prop.badgeTextGen(text)} icon={prop.icon} />;
+      }
+    });
+    return node;
+  });
+};
```
ivy-lli, line 64: a little sad, that you now have to run the regex multiple times.
but I also see currently no other solution...

ivy-lli (approved):

PR #180, opened 2024-10-24
XIVY-15045 Add initial selected row to browser

ivy-edp:
It should be possible to define whether a row is selected by default. This idea originated from the type browser in the DataClass editor; if a type is already present in the input, it should be selected upon the initial opening of the browser. This will help users better understand their current context within the browser.

### DIFF
--- a/packages/components/src/components/common/table/hooks/hooks.tsx
+++ b/packages/components/src/components/common/table/hooks/hooks.tsx
@@ -9,7 +9,8 @@ import {
   type TableState,
   getExpandedRowModel,
   type Row,
-  type Table
+  type Table,
+  type RowSelectionState
 } from '@tanstack/react-table';
 import * as React from 'react';
 
@@ -45,8 +46,8 @@ type UseTableSelectRetunValue<TData> = {
   tableState: Partial<TableState>;
 };
 
-export const useTableSelect = <TData,>(): UseTableSelectRetunValue<TData> => {
-  const [rowSelection, setRowSelection] = React.useState({});
+export const useTableSelect = <TData,>(initialSelecteState?: RowSelectionState): UseTableSelectRetunValue<TData> => {
+  const [rowSelection, setRowSelection] = React.useState(initialSelecteState ?? {});
   return {
     options: {
       onRowSelectionChange: setRowSelection,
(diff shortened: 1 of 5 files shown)


packages/components/src/components/common/table/hooks/hooks.tsx
```diff
@@ -45,8 +46,8 @@ type UseTableSelectRetunValue<TData> = {
   tableState: Partial<TableState>;
 };
 
-export const useTableSelect = <TData,>(): UseTableSelectRetunValue<TData> => {
-  const [rowSelection, setRowSelection] = React.useState({});
+export const useTableSelect = <TData,>(initialSelected?: RowSelectionState): UseTableSelectRetunValue<TData> => {
+  const [rowSelection, setRowSelection] = React.useState(initialSelected ? initialSelected : {});
```
ivy-lli:
```suggestion
  const [rowSelection, setRowSelection] = React.useState(initialSelected ?? {});
```

packages/components/src/components/editor/browser/browser.tsx
```diff
@@ -36,7 +37,8 @@ export const useBrowser = (
   data: Array<BrowserNode>,
   loadChildren?: (row: Row<BrowserNode>) => void,
   initialSearch?: string,
```
ivy-edp, line 39: Perhaps it now makes sense to create an "options" object instead of passing each option individually.

packages/components/src/components/editor/browser/browser.stories.tsx
```diff
@@ -30,7 +30,7 @@ type DefaultBrowserProps = {
 };
 
 const DefaultBrowser = ({ applyFn, applyBtn, initSearch }: DefaultBrowserProps) => {
-  const roles = useBrowser(roleData, undefined, initSearch);
+  const roles = useBrowser(roleData, undefined, initSearch, undefined, { '0.1': true });
```
ivy-lli, line 33: This is exactly the reason why at a specific point a option object makes sense.
So I know this means a breaking change, but I would not change this so all optional properties for this hook live in a optional object, so this gets possible
`useBrowser(roleData, {initialSearch: initSearch, initialSelected: { '0.1': true }});`

ivy-lli (changes requested):

packages/components/src/components/editor/browser/browser.tsx
ivy-lli: Yes look below, you were faster 😆

ivy-lli (approved):

PR #181, opened 2024-10-24
XIVY-14728 adds support for dynamic border

ivy-fhe:


packages/components/src/components/common/inputbadge/inputBadge.stories.tsx
```diff
@@ -44,4 +45,36 @@ export const BadgeArea: Story = {
...
+  )
+};
+
+export const InputBadgeError: Story = {
+  render: props => (
+    <BasicField message={{ message: 'this is an error', variant: 'error' }}>
+      <InputBadge {...props} />
+    </BasicField>
+  )
+};
```
ivy-lli, line 78: IMO not needed to add a example for each variant.
One would be enough for me, maybe with the default setting warning or error and then define a storybook argument so you can switch it in the view.
E.g. like the Message/WithInput does it

ivy-lli (approved):

PR #184, opened 2024-10-29
XIVY-15123 Enable placement of a Button in the top right corner of the Browser

ivy-edp:
Now I can add for example a ... button in the top right corner to enable additional functionality in the browser.
grafik

### DIFF
--- a/packages/components/src/components/editor/browser/browser.stories.tsx
+++ b/packages/components/src/components/editor/browser/browser.stories.tsx
@@ -8,6 +8,8 @@ import { Button } from '@/components/common/button/button';
 import { BasicCheckbox } from '@/components/common/checkbox/checkbox';
 import { Dialog, DialogContent, DialogHeader, DialogTitle, DialogTrigger } from '@/components/common/dialog/dialog';
 import { InputGroup, Input } from '@/components/common/input/input';
+import { Popover, PopoverArrow, PopoverContent, PopoverTrigger } from '@radix-ui/react-popover';
+import { Flex } from '@/components/common/flex/flex';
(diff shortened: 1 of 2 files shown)


ivy-lgi (approved):

packages/components/src/components/editor/browser/browser.tsx
```diff
@@ -94,6 +94,7 @@ export type Browser = {
   icon: IvyIcons;
   browser: ReturnType<typeof useBrowser>;
   header?: React.ReactNode;
+  headerControl?: React.ReactNode;
```
ivy-lli, line 97: why... 😥

We should not add and add and add properties. If we think like that the component will get more and more complicated! 
Beside of that case an adopter of this component need to know that the header and headerControl are both needed if he wants to add a headerControl. If not the outcome is not the same!

Why should the adapter not be able to write this Flex wrapper?!?

packages/components/src/components/editor/browser/browser.stories.tsx
```diff
@@ -65,6 +67,7 @@ const DefaultBrowser = ({ applyFn, applyBtn, initSearch }: DefaultBrowserProps)
           icon: IvyIcons.Cms,
           browser: cms,
           header: 'Info: More info content (with language details) / value modified with macro tags',
+          headerControl: QuickActionPopover(),
```
ivy-lli, line 70: This makes me sad 😭!

ivy-lli (commented):
Please no!
This is by the way the reason why I added myself as codeowner in this repo...

PR #185, opened 2024-10-30
XIVY-14728 add style and classname param

ivy-fhe:


ivy-edp (approved):

packages/components/src/components/common/inputbadge/inputBadge.tsx
```diff
@@ -10,6 +10,8 @@ import { splitNewLine } from '@/utils/string';
 export type InputBadgeProps = {
   value: string;
   badgeProps: Array<BadgeProps>;
+  style?: CSSProperties;
+  className?: string;
```
ivy-lli, line 14: Let me be clear this is not wrong.
But mostly all other components just reuse the props from the root element and just pass them through.
Why not just do this here too?

PR #186, opened 2024-11-04
XIVY-15123 Remove headerControl-Props from Browser-Component

ivy-edp:


packages/components/src/components/editor/browser/browser.stories.tsx
```diff
@@ -82,36 +81,39 @@ const DefaultBrowser = ({ applyFn, applyBtn, initSearch }: DefaultBrowserProps)
   );
 };
 
-const QuickActionPopover = () => {
+const QuickActionPopover = ({ additionalInfo }: { additionalInfo?: React.ReactNode }) => {
```
ivy-lli, line 84: As we don't know if we really do it this way, I would not call this `QuickActionPopover`.
And this here is an example what you can do.

I would just call it `CmsHeader`. 
And as it is just an example code you don't have to put the info text as property. Just write it in the component itself.

packages/components/src/components/editor/browser/browser.stories.tsx
```diff
@@ -81,12 +81,12 @@ const DefaultBrowser = ({ applyFn, applyBtn, initSearch }: DefaultBrowserProps)
...
 };
 
-const QuickActionPopover = ({ additionalInfo }: { additionalInfo?: React.ReactNode }) => {
+const CmsHeader = () => {
   const [popoverOpen, setPopoverOpen] = useState(false);
 
   return (
-    <Flex direction='row' justifyContent={additionalInfo ? 'space-between' : 'flex-end'} alignItems='center'>
-      {additionalInfo}
+    <Flex direction='row' justifyContent={'space-between'} alignItems='center'>
```
ivy-lli, line 88: if you have only a string the currly backets are not needed

ivy-lli (approved):
thx 👍

PR #192, opened 2024-11-07
Use only one pipeline

ivy-lli:
Sorry didn't get that in the first PR, but IMO we don't need a second pipeline for this right?

Maybe we want oneday also the font to be part of this maven zip then we need the node build output anyway...


ivy-rew (approved):

ivy-lgi (approved):
:+1:

PR #195, opened 2024-11-27
XIVY-15533 Add ConditionBuilder

ivy-edp:
Here is my suggestion for the Condition Builder in UI Components (currently just a draft, as I still want to add tests, but you can take a look at it in the meantime):
In the Browser story, I demonstrate how to integrate the Condition Builder as a browser into the BrowserView, and in the ConditionBuilder story, I show how to use the condition builder independently (so it can already be used in the inscription as well).

### DIFF
--- a/packages/components/src/components/editor/browser/browser.stories.tsx
+++ b/packages/components/src/components/editor/browser/browser.stories.tsx
@@ -10,6 +10,12 @@ import { Dialog, DialogContent, DialogHeader, DialogTitle, DialogTrigger } from 
 import { InputGroup, Input } from '@/components/common/input/input';
 import { Popover, PopoverArrow, PopoverContent, PopoverTrigger } from '@radix-ui/react-popover';
 import { Flex } from '@/components/common/flex/flex';
+import {
+  generateConditionString,
+  logicalOperatorOptions,
+  typeOptions
+} from '@/components/editor/condition-builder/condition-builder-utils';
+import { useConditionBuilder } from '@/components/editor/condition-builder/conditionBuilder';
 
 const meta: Meta<typeof BrowsersView> = {
   title: 'Editor/BrowsersView',
@@ -29,13 +35,21 @@ type DefaultBrowserProps = {
   applyFn?: (value?: string) => void;
   applyBtn?: { label?: string; icon?: IvyIcons };
   initSearch?: string;
+  includeConditionBuilder?: boolean;
 };
 
-const DefaultBrowser = ({ applyFn, applyBtn, initSearch }: DefaultBrowserProps) => {
+const DefaultBrowser = ({ applyFn, applyBtn, initSearch, includeConditionBuilder }: DefaultBrowserProps) => {
   const roles = useBrowser(roleData, { initialSearch: initSearch, initialSelecteState: { '0.1': true } });
   const attrs = useAttrBrowser();
   const funcs = useBrowser(funcData);
   const cms = useBrowser(cmsData);
+
+  const condition = useConditionBuilder({
+    generateConditionString: generateConditionString,
+    typeOptions,
+    logicalOperatorOptions,
+    argumentInput: (value, onChange) => <DialogBrowserInput value={value} onChange={onChange} />
+  });
 
   return (
     <BrowsersView
@@ -69,7 +83,8 @@ const DefaultBrowser = ({ applyFn, applyBtn, initSearch }: DefaultBrowserProps) 
           header: <CmsHeader />,
           infoProvider: row => <CmsInfoProvider row={row} />,
           applyModifier: row => ({ value: `<%= ivy.co('${row?.original.value}') %>` })
-        }
+        },
+        ...(includeConditionBuilder ? [condition] : [])
       ]}
       apply={(browserName, result) => {
         console.log('apply', browserName, result);
@@ -118,32 +133,45 @@ const CmsHeader = () => {
 };
 
 export const Default: Story = {
-  render: () => <DefaultBrowser />
+  render: () => <DefaultBrowser includeConditionBuilder={true} />
+};
+
+const DialogBrowserInput = ({
+  value,
+  onChange,
+  includeConditionBuilder
+}: {
+  value: string;
+  onChange: (change: string) => void;
+  includeConditionBuilder?: boolean;
+}) => {
+  const [open, setOpen] = useState(false);
+  return (
+    <Dialog open={open} onOpenChange={setOpen}>
+      <InputGroup style={{ flex: '1' }}>
+        <Input value={value} onChange={e => onChange(e.target.value)} />
+        <DialogTrigger asChild>
+          <Button icon={IvyIcons.ListSearch} aria-label='Browser' />
+        </DialogTrigger>
+      </InputGroup>
+      <DialogContent style={{ height: '80vh' }}>
+        <DefaultBrowser
+          applyFn={value => {
+            if (value) onChange(value);
+            setOpen(false);
+          }}
+          initSearch={value}
+          includeConditionBuilder={includeConditionBuilder}
+        />
+      </DialogContent>
+    </Dialog>
+  );
 };
 
 export const DialogBrowser: Story = {
   render: () => {
-    const [open, setOpen] = useState(false);
     const [input, setInput] = useState('');
-    return (
-      <Dialog open={open} onOpenChange={setOpen}>
-        <InputGroup>
-          <Input value={input} onChange={e => setInput(e.target.value)} />
-          <DialogTrigger asChild>
-            <Button icon={IvyIcons.ListSearch} aria-label='Browser' />
-          </DialogTrigger>
-        </InputGroup>
-        <DialogContent style={{ height: '80vh' }}>
-          <DefaultBrowser
-            applyFn={value => {
-              if (value) setInput(value);
-              setOpen(false);
-            }}
-            initSearch={input}
-          />
-        </DialogContent>
-      </Dialog>
-    );
+    return <DialogBrowserInput value={input} onChange={setInput} />;
   }
 };
 
--- a/packages/components/src/components/editor/condition-builder/condition.tsx
+++ b/packages/components/src/components/editor/condition-builder/condition.tsx
@@ -1,0 +1,57 @@
+import { Button } from '@/components/common/button/button';
+import { BasicField } from '@/components/common/field/field';
+import { Flex } from '@/components/common/flex/flex';
+import { BasicSelect } from '@/components/common/select/select';
+import { conditionBorder, logicalOperatorField, operatorField } from '@/components/editor/condition-builder/conditionBuilder.css';
+import {
+  useConditionBuilderContext,
+  type Operator,
+  type LogicOperator,
+  type ConditionData
+} from '@/components/editor/condition-builder/conditionBuilderContext';
+import { cn } from '@/utils/class-name';
+import { IvyIcons } from '@axonivy/ui-icons';
+import { Label } from '@radix-ui/react-label';
+
+export interface ConditionProps {
+  condition: ConditionData;
+  conditionIndex: number;
+  groupIndex: number;
+  conditionsCount: number;
+}
+
+export const Condition = ({ condition, conditionIndex, groupIndex, conditionsCount }: ConditionProps) => {
+  const { updateCondition, removeCondition, typeOptions, logicalOperatorOptions, argumentInput } = useConditionBuilderContext();
+  return (
+    <Flex direction='column' className={cn(conditionBorder, 'ui-condition-builder-condition')}>
+      <Flex direction='row' justifyContent='space-between'>
+        <Label>{`Condition ${conditionIndex + 1}`}</Label>
+        <Button onClick={() => removeCondition(groupIndex, conditionIndex)} icon={IvyIcons.Trash} aria-label='Remove Condition' />
+      </Flex>
+      <Flex alignItems='center' gap={2} direction='row'>
+        {argumentInput(condition.argument1, val => updateCondition(groupIndex, conditionIndex, 'argument1', val))}
+        <BasicField className={operatorField}>
+          <BasicSelect
+            items={typeOptions}
+            value={condition.operator}
+            onValueChange={val => updateCondition(groupIndex, conditionIndex, 'operator', val as Operator)}
+          />
+        </BasicField>
+        {condition.operator !== 'is true' &&
+          condition.operator !== 'is false' &&
+          condition.operator !== 'is empty' &&
+          condition.operator !== 'is not empty' &&
+          argumentInput(condition.argument2, val => updateCondition(groupIndex, conditionIndex, 'argument2', val))}
+        {conditionIndex < conditionsCount - 1 && (
+          <BasicField className={logicalOperatorField}>
+            <BasicSelect
+              items={logicalOperatorOptions}
+              value={condition.logicalOperator}
+              onValueChange={val => updateCondition(groupIndex, conditionIndex, 'logicalOperator', val as LogicOperator)}
+            />
+          </BasicField>
+        )}
+      </Flex>
+    </Flex>
+  );
+};
--- a/packages/components/src/components/editor/condition-builder/conditionBuilder.css.ts
+++ b/packages/components/src/components/editor/condition-builder/conditionBuilder.css.ts
@@ -1,0 +1,24 @@
+import { vars } from '@/styles/theme.css';
+import { style } from '@vanilla-extract/css';
+
+const sharedConditionStyles = {
+  border: vars.border.basic,
+  borderRadius: vars.border.r2,
+  padding: vars.size.s2
+};
+
+export const conditionBorder = style(sharedConditionStyles);
+
+export const conditionGroupBorder = style({
+  selectors: {
+    '&[data-condition-mode="nested-condition"]': sharedConditionStyles
+  }
+});
+
+export const operatorField = style({
+  width: 150
+});
+
+export const logicalOperatorField = style({
+  width: 70
+});
--- a/packages/components/src/components/editor/condition-builder/conditionBuilder.stories.tsx
+++ b/packages/components/src/components/editor/condition-builder/conditionBuilder.stories.tsx
@@ -1,0 +1,39 @@
+import type { Meta, StoryObj } from '@storybook/react';
+import { useState } from 'react';
+import {
+  generateConditionString,
+  logicalOperatorOptions,
+  typeOptions
+} from '@/components/editor/condition-builder/condition-builder-utils';
+import { ConditionBuilderProvider } from '@/components/editor/condition-builder/conditionBuilderContext';
+import { ConditionBuilder } from '@/components/editor/condition-builder/conditionBuilder';
+import { vars } from '@/styles/theme.css';
+
+const meta: Meta<typeof ConditionBuilder> = {
+  title: 'Editor/ConditionBuilder',
+  component: ConditionBuilder
+};
+
+export default meta;
+
+type Story = StoryObj<typeof ConditionBuilder>;
+
+export const Default: Story = {
+  render: () => {
+    const [condition, setCondition] = useState<string>();
+    return (
+      <>
+        <ConditionBuilderProvider
+          generateConditionString={generateConditionString}
+          logicalOperatorOptions={logicalOperatorOptions}
+          typeOptions={typeOptions}
+        >
+          <ConditionBuilder onChange={setCondition} />
+        </ConditionBuilderProvider>
+        <pre style={{ border: vars.border.basic, maxHeight: '80px', padding: vars.padding.input, borderRadius: vars.border.r2 }}>
+          {condition}
+        </pre>
+      </>
+    );
+  }
+};
--- a/packages/components/src/components/editor/condition-builder/conditionBuilder.tsx
+++ b/packages/components/src/components/editor/condition-builder/conditionBuilder.tsx
@@ -1,0 +1,78 @@
+import { IvyIcons } from '@axonivy/ui-icons';
+import { BasicSelect } from '@/components/common/select/select';
+import { Flex } from '@/components/common/flex/flex';
+import { Button } from '@/components/common/button/button';
+import { useEffect, useState } from 'react';
+import type { Browser } from '@/components/editor/browser/browser';
+import {
+  ConditionBuilderProvider,
+  useConditionBuilderContext,
+  type ConditionBuilderProviderProps
+} from '@/components/editor/condition-builder/conditionBuilderContext';
+import { ConditionGroup } from './conditionGroup';
+
+interface ConditionBuilderProps {
+  onChange: (value: string) => void;
+}
+
+type ConditionBuilderHookProps = Omit<ConditionBuilderProviderProps, 'children'>;
+
+const conditionModes = [
+  { label: 'Basic Condition', value: 'basic-condition' },
+  { label: 'Nested Condition', value: 'nested-condition' },
+  { label: 'Always True', value: 'always-true' },
+  { label: 'Always False', value: 'always-false' }
+];
+export type ConditionMode = (typeof conditionModes)[number]['value'];
+
+export const useConditionBuilder = (props: ConditionBuilderHookProps): Browser => {
+  const [value, setValue] = useState<string>();
+
+  return {
+    name: 'Condition',
+    browser: (
+      <ConditionBuilderProvider {...props}>
+        <ConditionBuilder onChange={setValue} />
+      </ConditionBuilderProvider>
+    ),
+    applyModifier: () => {
+      return { value: value || '' };
+    },
+    infoProvider: () => <p>{value}</p>,
+    icon: IvyIcons.Process
+  };
+};
+
+const ConditionBuilder = ({ onChange }: ConditionBuilderProps) => {
+  const { addConditionGroup, generateConditionString, conditionGroups, conditionMode, setConditionMode } = useConditionBuilderContext();
+
+  useEffect(() => {
+    onChange(generateConditionString());
+  }, [generateConditionString, onChange]);
+
+  return (
+    <Flex direction='column' gap={2}>
+      <BasicSelect items={conditionModes} value={conditionMode} onValueChange={val => setConditionMode(val as ConditionMode)} />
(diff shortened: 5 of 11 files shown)


packages/components/src/components/editor/condition-builder/ConditionBuilderContext.tsx
```diff
@@ -0,0 +1,155 @@
...
+  'is false': string;
+  'is empty': string;
+  'is not empty': string;
+  'less than': string;
+  'greater than': string;
+  'less or equal to': string;
+  'greater or equal to': string;
+};
+
+export type ConditionType = keyof TypeOperator;
```
ivy-edp, line 20: Unfortunately, I haven't found a way to define a type that takes all the values from, for example, typesOptions (since the typesOptions object can't be accessed at the time the type is generated). So i just use the keyof TypeOperator. Also not quiet sure about the naming of TypeOperator/ConditionType and LogicOperator/LogicalOperator

packages/components/src/components/editor/condition-builder/conditionBuilder.stories.tsx
```diff
@@ -0,0 +1,110 @@
...
+    case 'is true':
+      return `${formattedArg1}${logicalOp}`;
+    case 'is false':
+      return `!${formattedArg1}${logicalOp}`;
+    case 'is empty':
+      return `${formattedArg1}.isEmpty()${logicalOp}`;
+    case 'is not empty':
+      return `!${formattedArg1}.isEmpty()${logicalOp}`;
+    default:
+      return `${formattedArg1} ${TypeOptions[con.operator]} ${formattedArg2}${logicalOp}`;
```
ivy-edp, line 108: Here you can see that as a consequence that ConditionType directly uses keyof TypeOperator and not the actual value of the attribute, I need to search for the correct value in the implementation (if it's unclear, we can discuss it in person)

packages/components/src/components/editor/condition-builder/conditionBuilder.stories.tsx
```diff
@@ -0,0 +1,110 @@
...
+  'less or equal to': '<=',
+  'greater or equal to': '>='
+};
+
+export const LogicalOperatorOptions: LogicOperator = {
+  'and': '&&',
+  'or': '||'
+};
+
+export const GenerateConditionString = (conditionMode: ConditionMode, conditionGroups: ConditionGroup[]) => {
```
ivy-edp, line 62: I'm not sure if it's a good idea to define these const (TypeOptions, LogicalOperatorOptions, GeneratedConditionString) as exports. I did it because this allows me to easily reuse them in the Browser story.

packages/components/src/components/editor/browser/browser.stories.tsx
```diff
@@ -37,6 +43,16 @@ const DefaultBrowser = ({ applyFn, applyBtn, initSearch }: DefaultBrowserProps)
...
 
+  const argumentInput = (value: string, onChange: (change: string) => void) => {
+    return <DialogBrowserInput value={value} onChange={onChange} />;
+  };
+  const condition = useConditionBuilder({
+    generateConditionString: GenerateConditionString,
+    typeOptions: TypeOptions,
+    logicalOperatorOptions: LogicalOperatorOptions,
+    argumentInput
+  });
```
ivy-lli, line 54: I think I would directly include the function instead of define it above

```suggestion
  const argumentInput = (value: string, onChange: (change: string) => void) => {
    return <DialogBrowserInput value={value} onChange={onChange} />;
  };
  const condition = useConditionBuilder({
    generateConditionString: GenerateConditionString,
    typeOptions: TypeOptions,
    logicalOperatorOptions: LogicalOperatorOptions,
    argumentInput: (value, onChange) => <DialogBrowserInput value={value} onChange={onChange} />
  });
```

packages/components/src/components/editor/browser/browser.stories.tsx
```diff
@@ -121,29 +138,33 @@ export const Default: Story = {
...
+  return (
+    <Dialog open={open} onOpenChange={setOpen}>
+      <InputGroup style={{ flex: '1' }}>
+        <Input value={value} onChange={e => onChange(e.target.value)} />
+        <DialogTrigger asChild>
+          <Button icon={IvyIcons.ListSearch} aria-label='Browser' />
+        </DialogTrigger>
+      </InputGroup>
+      <DialogContent style={{ height: '80vh' }}>
+        <DefaultBrowser
```
ivy-lli, line 152: Isn't it possible to open in this browser also a condition builder? 🙈

packages/components/src/components/editor/condition-builder/conditionBuilder.stories.tsx
```diff
@@ -0,0 +1,110 @@
...
+          style={{ border: 'var(--basic-border)', maxHeight: '80px', padding: 'var(--input-padding)', borderRadius: 'var(--border-r2)' }}
+        >
+          {condition}
+        </pre>
+      </>
+    );
+  }
+};
+
+export const TypeOptions: TypeOperator = {
```
ivy-lli, line 44: this is an object and should not start uppercase
```suggestion
export const typeOptions: TypeOperator = {
```

packages/components/src/components/editor/condition-builder/conditionBuilder.stories.tsx
```diff
@@ -0,0 +1,110 @@
...
+  'is false': 'isFalse',
+  'is empty': 'isEmpty',
+  'is not empty': 'isNotEmpty',
+  'less than': '<',
+  'greater than': '>',
+  'less or equal to': '<=',
+  'greater or equal to': '>='
+};
+
+export const LogicalOperatorOptions: LogicOperator = {
```
ivy-lli, line 57: this is an object and should not start uppercase
```suggestion
export const logicalOperatorOptions: LogicOperator = {
```

packages/components/src/components/editor/condition-builder/conditionBuilder.stories.tsx
ivy-lli: I think this is fine. 
This stories are not build in the library bundle output. They only live in the storybook output.

ivy-lli: But this is a function and should not be uppercase
```suggestion
export const generateConditionString = (conditionMode: ConditionMode, conditionGroups: ConditionGroup[]) => {
```

packages/components/src/components/editor/condition-builder/conditionBuilder.stories.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+  return groupStrings.join('');
+};
+
+const getLogicalOperator = (index: number, totalConditions: number, logicalOperator: LogicalOperator) => {
+  return index < totalConditions - 1 ? ` ${LogicalOperatorOptions[logicalOperator].toLowerCase()} ` : '';
+};
+
+const formatCondition = (con: Condition, formattedArg1: string, formattedArg2: string, logicalOp: string) => {
+  switch (con.operator) {
```
ivy-lli, line 98: con is not a very good name IMO. I would call it condition

packages/components/src/components/editor/condition-builder/Condition.tsx
```diff
@@ -0,0 +1,75 @@
...
+import { BasicSelect } from '@/components/common/select/select';
+import {
+  useConditionBuilderContext,
+  type ConditionType,
+  type LogicalOperator
+} from '@/components/editor/condition-builder/ConditionBuilderContext';
+import { IvyIcons } from '@axonivy/ui-icons';
+import { Label } from '@radix-ui/react-label';
+
+export interface Condition {
```
ivy-lli, line 13: I think it is not idea that the type here has the same name as the Component.

packages/components/src/components/editor/condition-builder/conditionBuilder.stories.tsx
ivy-lli: I'm not sure what you mean, but this is fine or not?
the con.operator is one of the keys of the TypeOperator, and the value for this key is the thing which is different and the thing we need to put here.

packages/components/src/components/editor/condition-builder/ConditionGroup.tsx
```diff
@@ -0,0 +1,66 @@
...
+interface ConditionGroupProps extends Pick<ConditionProps, 'groupIndex'> {
+  group: ConditionGroup;
+  groupCount: number;
+}
+
+export const ConditionGroup = ({ group, groupIndex, groupCount }: ConditionGroupProps) => {
+  const { updateLogicalOperator, addCondition, removeConditionGroup, conditionMode, logicalOperatorOptions } = useConditionBuilderContext();
+
+  return (
+    <Flex direction='column' gap={2} className='condition-builder__group'>
```
ivy-lli, line 23: I assume you have this class here also for the impls so they can test this more easily?
I started here to streamline this additional class names a bit (see the other components). 
They always start with `ui-` it think, so in this case I would name them something like `ui-condition-builder-group`

packages/components/src/components/editor/condition-builder/ConditionGroup.tsx
```diff
@@ -0,0 +1,66 @@
...
+      <Flex
+        direction='column'
+        style={
+          conditionMode === 'nested-condition'
+            ? { border: 'var(--basic-border', borderRadius: 'var(--border-r2)', padding: 'var(--size-2)' }
+            : undefined
+        }
```
ivy-lli, line 30: I think it is sometimes fine to use inline styles in the impls, but here I think we should do all the same way, via a separate CSS file, and use the known variables for this variables :)
See the other components for how.

And stuff like the condition mode should be a simple data argument, so you can adjust the styles in the CSS file according to it:
`data-condition-mode={conditionMode}`

packages/components/src/components/editor/condition-builder/ConditionGroup.tsx
```diff
@@ -0,0 +1,66 @@
...
+          />
+        ))}
+        <Button onClick={() => addCondition(groupIndex)} icon={IvyIcons.Plus} aria-label='Add Condition' variant='outline'>
+          Add Condition
+        </Button>
+      </Flex>
+      {groupIndex < groupCount - 1 && conditionMode === 'nested-condition' && (
+        <BasicSelect
+          items={Object.entries(logicalOperatorOptions)
+            .slice(0, -1)
```
ivy-lli, line 55: remove empty one?

packages/components/src/components/editor/condition-builder/ConditionGroup.tsx
```diff
@@ -0,0 +1,66 @@
...
+            conditionsCount={group.conditions.length}
+          />
+        ))}
+        <Button onClick={() => addCondition(groupIndex)} icon={IvyIcons.Plus} aria-label='Add Condition' variant='outline'>
+          Add Condition
+        </Button>
+      </Flex>
+      {groupIndex < groupCount - 1 && conditionMode === 'nested-condition' && (
+        <BasicSelect
+          items={Object.entries(logicalOperatorOptions)
```
ivy-lli, line 54: this options to items mapping is very global, and should be only once IMO, so not on every rerender.
E.g. in a useMemo at the start of the component.

packages/components/src/components/editor/condition-builder/ConditionBuilder.tsx
```diff
@@ -0,0 +1,72 @@
...
+}
+
+type ConditionBuilderHookProps = Omit<ConditionBuilderProviderProps, 'children'>;
+
+const conditionModes = [
+  { label: 'Basic Condition', value: 'basic-condition' },
+  { label: 'Nested Condition', value: 'nested-condition' },
+  { label: 'Always True', value: 'always-true' },
+  { label: 'Always False', value: 'always-false' }
+];
```
ivy-lli, line 21: best mark this as const (readonly)
```suggestion
const conditionModes = [
  { label: 'Basic Condition', value: 'basic-condition' },
  { label: 'Nested Condition', value: 'nested-condition' },
  { label: 'Always True', value: 'always-true' },
  { label: 'Always False', value: 'always-false' }
] as const;
```

packages/components/src/components/editor/condition-builder/ConditionBuilder.tsx
```diff
@@ -0,0 +1,72 @@
...
+    icon: IvyIcons.Process
+  };
+};
+
+export const ConditionBuilder = ({ onChange }: ConditionBuilderProps) => {
+  const { addConditionGroup, generateConditionString, conditionGroups, conditionMode, setConditionMode } = useConditionBuilderContext();
+
+  useEffect(() => {
+    onChange(generateConditionString());
+  }, [generateConditionString, onChange]);
```
ivy-lli, line 47: why is this needed?

ivy-lli (commented):
not finished, but here some first feedback.

packages/components/src/components/editor/condition-builder/conditionBuilder.stories.tsx
ivy-edp: I capitalised these objects because otherwise I got an ESLint error: https://jenkins.ivyteam.io/job/ui-components/job/add-condition-builder-to-browser/4/eslint/

packages/components/src/components/editor/condition-builder/ConditionGroup.tsx
ivy-edp: No for some reason if i do it with Object.entries() the last array is always "parameters" no idea why, so i remove it.
grafik

packages/components/src/components/editor/condition-builder/conditionBuilder.stories.tsx
ivy-lli: Ahh because an export in a story is expected as a story in which you then can see in the storybook ui. 
Ok then best define this things in an additional file (like it is done in the table stories I think).
As long as you don't export it in the toplevel index, it will not be included in the bundle output

packages/components/src/components/editor/condition-builder/ConditionGroup.tsx
ivy-edp: Or what do you think of the idea if I simply do the mapping of logicalOperatorOptions and typeOptions directly in the useConditionBuilderContext hook and directly return a corresponding array of label/value objects?

packages/components/src/components/editor/condition-builder/ConditionGroup.tsx
ivy-lli: seems like a hack to me. we should better know why this is and fix it at the source :)

packages/components/src/components/editor/condition-builder/ConditionBuilder.tsx
ivy-edp: This is needed to ensure that the Info-Collapsible displays the current condition at all times. The info is defined outside the builder and can be edited via onChange.

packages/components/src/components/editor/browser/browser.stories.tsx
ivy-edp: Oh yes, that's right 😄, I can add a flag that the condition builder is not integrated for this case.

packages/components/src/components/editor/condition-builder/ConditionGroup.tsx
ivy-edp: Yes I agree 😅 However, for some reason after I moved the hole Object.entries().map logic into the ConditionBuilderContext, there was no longer the issue with this parameters and i was able to remove the slice 👍

packages/components/src/components/editor/condition-builder/ConditionBuilder.tsx
ivy-lli: ah okey. could be improved maybe :) 
leave it for now :)

packages/components/src/components/editor/condition-builder/Condition.tsx
```diff
@@ -0,0 +1,75 @@
...
+  groupIndex: number;
+  conditionsCount: number;
+}
+
+export const Condition = ({ condition, conditionIndex, groupIndex, conditionsCount }: ConditionProps) => {
+  const { updateCondition, removeCondition, typeOptions, logicalOperatorOptions, argumentInput } = useConditionBuilderContext();
+  return (
+    <Flex
+      direction='column'
+      style={{ border: 'var(--basic-border', borderRadius: 'var(--border-r2)', padding: 'var(--size-2)' }}
```
ivy-lli, line 32: move to css file

packages/components/src/components/editor/condition-builder/Condition.tsx
```diff
@@ -0,0 +1,75 @@
...
+      className='condition-builder__condition'
```
ivy-lli, line 33: streamline with other classnames: `ui-condition-builder-condition`

packages/components/src/components/editor/condition-builder/ConditionBuilderContext.tsx
ivy-lli: It is not possible to generate a type of something which is not constant or already an type
But why would you need that, just make a type of the static part of the type and use it to access the value for that key.

ivy-lli: not sure too... maybe:

TypeOperator -> OperatorType (or just Operators)
ConditionType -> Operator
LogicOperator -> LogicOperators

I think I would rename the ConditionType, as this sais nothing right? it is an operator not a condition type...

packages/components/src/components/editor/condition-builder/ConditionBuilderContext.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+export type LogicOperator = {
+  'and': string;
+  'or': string;
+};
+
+export type LogicalOperator = keyof LogicOperator;
```
ivy-lli, line 27: use one of them but not both :D
I would streamline this name with the "normal" operator type

packages/components/src/components/editor/condition-builder/ConditionBuilder.tsx
ivy-lli: maybe I'm wrong, but this looks like this will be executed a lot... At least multiple times per change in the builder.
Because when does the useEffect here know that something has changed inside?

packages/components/src/components/editor/condition-builder/ConditionBuilderContext.tsx
ivy-edp: I renamed the types no like this, what do you think?
grafik

packages/components/src/components/editor/condition-builder/Condition.tsx
ivy-edp: i renamed the Condition interface and the ConditionGroup interface to "ConditionData" and "ConditionGroupData". What do you think?

packages/components/src/components/editor/condition-builder/condition.tsx
```diff
@@ -0,0 +1,57 @@
...
+  const { updateCondition, removeCondition, typeOptions, logicalOperatorOptions, argumentInput } = useConditionBuilderContext();
+  return (
+    <Flex direction='column' className={cn(conditionBorder, 'ui-condition-builder-condition')}>
+      <Flex direction='row' justifyContent='space-between'>
+        <Label>{`Condition ${conditionIndex + 1}`}</Label>
+        <Button onClick={() => removeCondition(groupIndex, conditionIndex)} icon={IvyIcons.Trash} aria-label='Remove Condition' />
+      </Flex>
+      <Flex alignItems='center' gap={2} direction='row'>
+        {argumentInput(condition.argument1, val => updateCondition(groupIndex, conditionIndex, 'argument1', val))}
+        <BasicField style={{ width: '150px' }}>
```
ivy-lli, line 33: maybe move this to the css file?

packages/components/src/components/editor/condition-builder/condition.tsx
```diff
@@ -0,0 +1,57 @@
...
+            onValueChange={val => updateCondition(groupIndex, conditionIndex, 'operator', val as Operator)}
+          />
+        </BasicField>
+        {condition.operator !== 'is true' &&
+          condition.operator !== 'is false' &&
+          condition.operator !== 'is empty' &&
+          condition.operator !== 'is not empty' &&
+          argumentInput(condition.argument2, val => updateCondition(groupIndex, conditionIndex, 'argument2', val))}
+        {conditionIndex < conditionsCount - 1 && (
+          <BasicField style={{ width: '70px' }}>
```
ivy-lli, line 46: maybe move to the css file?

packages/components/src/components/editor/condition-builder/conditionBuilder.stories.tsx
```diff
@@ -0,0 +1,40 @@
...
+      <>
+        <ConditionBuilderProvider
+          generateConditionString={generateConditionString}
+          logicalOperatorOptions={logicalOperatorOptions}
+          typeOptions={typeOptions}
+        >
+          <ConditionBuilder onChange={setCondition} />
+        </ConditionBuilderProvider>
+        <pre
+          style={{ border: 'var(--basic-border)', maxHeight: '80px', padding: 'var(--input-padding)', borderRadius: 'var(--border-r2)' }}
```
ivy-lli, line 33: maybe use the known variables?
```suggestion
          style={{ border: vars.border.basic, maxHeight: '80px', padding: vars..., borderRadius: vars.border.r2' }}
```

packages/components/src/components/editor/condition-builder/conditionGroup.tsx
```diff
@@ -0,0 +1,53 @@
...
+interface ConditionGroupProps extends Pick<ConditionProps, 'groupIndex'> {
+  group: ConditionGroupData;
+  groupCount: number;
+}
+
+export const ConditionGroup = ({ group, groupIndex, groupCount }: ConditionGroupProps) => {
+  const { updateLogicalOperator, addCondition, removeConditionGroup, conditionMode, logicalOperatorOptions } = useConditionBuilderContext();
+  return (
+    <Flex direction='column' gap={2} className='ui-condition-builder-group' data-condition-mode={conditionMode}>
+      <Flex direction='column' className={conditionMode === 'nested-condition' ? cn(conditionBorder) : ''} gap={2}>
```
ivy-lli, line 24: didn't you check what `cn` does?
```suggestion
      <Flex direction='column' className={cn(conditionBorder, conditionMode === 'nested-condition' && 'nested-condition')} gap={2}>
```

or why not add this mode as data arg? sounds more like a data arg than a classname...
```suggestion
      <Flex direction='column' className={conditionBorder} data-condition-mode={conditionMode} gap={2}>
```

packages/components/src/components/editor/condition-builder/conditionBuilder.tsx
```diff
@@ -0,0 +1,80 @@
...
+      </ConditionBuilderProvider>
+    ),
+    applyModifier: () => {
+      return { value: value || '' };
+    },
+    infoProvider: () => <p>{value}</p>,
+    icon: IvyIcons.Process
+  };
+};
+useConditionBuilder.displayName = 'useConditionBuilder';
```
ivy-lli, line 45: why do you add a displayName to a hook? I think this makes no sense or am I wrong?
The displayName is a react thing to name components. E.g. they will be shown like that in the react dev tools.
So it is nice if the components have such a displayName, but I think this is useless on hooks.

packages/components/src/components/editor/condition-builder/conditionBuilder.tsx
```diff
@@ -0,0 +1,80 @@
...
+    <Flex direction='column' gap={2}>
+      <BasicSelect items={conditionModes} value={conditionMode} onValueChange={val => setConditionMode(val as ConditionMode)} />
+      {conditionMode !== 'always-true' &&
+        conditionMode !== 'always-false' &&
+        conditionGroups
+          .map((group, groupIndex) => {
+            if (conditionMode === 'basic-condition' && groupIndex > 0) return null;
+            return <ConditionGroup key={groupIndex} group={group} groupIndex={groupIndex} groupCount={conditionGroups.length} />;
+          })
+          .filter(Boolean)}
```
ivy-lli, line 64: you can also filter first or not? I think then also the filter at the end would not be needed...

```suggestion
          .filter((group, groupIndex) => !(conditionMode === 'basic-condition' && groupIndex > 0))
          .map((group, groupIndex) => <ConditionGroup key={groupIndex} group={group} groupIndex={groupIndex} groupCount={conditionGroups.length} />)
```

packages/components/src/components/editor/condition-builder/conditionBuilderContext.tsx
```diff
@@ -0,0 +1,172 @@
...
+  and: string;
+  or: string;
+};
+
+export type LogicOperator = keyof LogicOperators;
+
+export interface ConditionBuilderProviderProps {
+  typeOptions: Operators;
+  logicalOperatorOptions: LogicOperators;
+  generateConditionString: (conditionMode: ConditionMode, conditionGroups: ConditionGroupData[]) => string;
```
ivy-lli, line 42: I prefer the Array<ConditionGroupData> syntax over the ConditionGroupData[] syntax...

ivy-edp: So should I change all the type definitions where i put bla[] instead of Array<bla>?

ivy-lli: If you want. I just wanted to mention it. 
https://www.totaltypescript.com/array-types-in-typescript

packages/components/src/components/editor/condition-builder/conditionBuilder.css.ts
```diff
@@ -1,8 +1,24 @@
...
+export const conditionBorder = style(sharedConditionStyles);
+
+export const conditionGroupBorder = style({
+  selectors: {
+    '&[data-condition-mode="nested-condition"]': sharedConditionStyles
+  }
+});
+
+export const operatorField = style({
+  width: '150px'
```
ivy-lli, line 19: just as info, if you have px then you can also write the following if you like:
```suggestion
  width: 150
```

ivy-lli (approved):
Looks good to me.
Thanks for the improvements 👍

PR #198, opened 2024-11-29
XIVY-15533 Enable browser applyModifier to not pass a row

ivy-edp:
To ensure that a applyModifier does not have to pass a row (which is the case for the ConditionBuilder, it is not a table), I made the row optional.


packages/components/src/components/editor/browser/browser.tsx
```diff
@@ -132,9 +132,16 @@ const BrowsersView = ({ browsers, apply, applyBtn }: BrowsersViewProps) => {
   const applyHandler = (doubleClickRow?: Row<BrowserNode>) => {
     const row = doubleClickRow ?? selectedRow();
     const browser = browsers.find(b => b.name === tab);
-    if (!row || !browser) {
+    if (!browser) {
       return; // nothing selected
```
ivy-lli, line 136: this comment is not longer true

packages/components/src/components/editor/browser/browser.tsx
```diff
@@ -132,9 +132,16 @@ const BrowsersView = ({ browsers, apply, applyBtn }: BrowsersViewProps) => {
...
     }
+    if (browser.applyModifier !== undefined && !row) {
+      apply(browser.name, browser.applyModifier());
+      return;
+    }
+    if (!row) {
+      return;
+    }
```
ivy-lli, line 144: instead to check the row twice I would do it like this:
```suggestion
    if (!row) {
      if (browser.applyModifier !== undefined) {
        apply(browser.name, browser.applyModifier());
      }
      return;
    }
```

ivy-lli (approved):

PR #202, opened 2024-12-06
Fix undefined class in rendered badge output

ivy-lli:


packages/components/src/components/common/inputbadge/inputBadge.tsx
```diff
@@ -23,24 +22,22 @@ const Badge = ({ icon, text }: { icon: IvyIcons; text: string }) => (
   </Flex>
 );
 
-export const InputBadge = forwardRef<ElementRef<'output'>, InputBadgeProps>(
-  ({ value, badgeProps, className, style, ...props }, forwardRef) => {
-    const { inputProps } = useField();
-    const items = useMemo(() => findBadges(value, badgeProps), [value, badgeProps]);
-    return (
-      <output className={`${inputBadgeOutput} ${className}`} style={style} {...inputProps} {...props} ref={forwardRef}>
```
ivy-lli, line 31: if you do it like this and className is undefined, undefined gets added as string to the dom
use the existing cn helper function for stuff like that

ivy-lli, line 31: there is no need to extra set the style prop, it is already part of props

ivy-fhe: :+1: good to know

ivy-fhe (approved):

PR #203, opened 2024-12-06
XIVY-15526 changes tooltip content to be only computed once hovered

ivy-fhe:
- only computes tooltipcontent once it is opened to minimize computations

### DIFF
--- a/packages/components/src/components/common/inputbadge/inputBadge.tsx
+++ b/packages/components/src/components/common/inputbadge/inputBadge.tsx
@@ -36,14 +36,15 @@ const Badge = ({ icon, text, originalText, tooltipGen }: BadgeProps) => (
           {text}
         </Flex>
       </TooltipTrigger>
-      {tooltipGen && (
-        <TooltipContent collisionPadding={10} sideOffset={10}>
-          {tooltipGen(originalText)}
-        </TooltipContent>
-      )}
+      <TooltipContent collisionPadding={10} sideOffset={10}>
+        <BadgeTooltip originalText={originalText} tooltipGen={tooltipGen} />
+      </TooltipContent>
     </Tooltip>
   </TooltipProvider>
 );
+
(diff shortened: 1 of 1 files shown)


packages/components/src/components/common/inputbadge/inputBadge.tsx
```diff
@@ -27,23 +27,26 @@ type BadgeProps = {
...
+        </TooltipTrigger>
+        {tooltipState && tooltipGen && (
+          <TooltipContent collisionPadding={10} sideOffset={10}>
+            {tooltipGen(originalText)}
+          </TooltipContent>
+        )}
+      </Tooltip>
+    </TooltipProvider>
+  );
+};
```
ivy-lli, line 49: other which should also work and I personally prefer would be move the call of the function to it's own component. Because the radix-ui tooltip content is only rendered when it will be shown (if it is a react component), if it is like it was before (direct call) then it will be resolved directly. By the way I think it would be good if the original text is show when no tooltipGen function is given. 
```suggestion
const Badge = ({ icon, text, ...props }: BadgeProps) => (
  <TooltipProvider>
    <Tooltip onOpenChange={setTooltipState}>
      <TooltipTrigger asChild>
        <Flex gap={1} alignItems='center' className={inputBadge}>
          <IvyIcon className={inputBadgeIcon} icon={icon} />
          {text}
        </Flex>
      </TooltipTrigger>
      <TooltipContent collisionPadding={10} sideOffset={10}>
        <BadgeTooltip {...props} />
      </TooltipContent>
    </Tooltip>
  </TooltipProvider>
);

const BadgeTooltip = ({originalText, tooltipGen}: Pick<BadgeProps, 'originalText' | 'tooltipGen'>}) => tooltipGen ? tooltipGen(originalText) : originalText;
```

ivy-fhe: :+1:

ivy-fhe:
> works right?

yes, works perfectly :smile:

PR #204, opened 2024-12-09
XIVY-15021 Add TableKeyHandler-Hook

ivy-edp:
With this new useTableKeyHandler hook, you are able to add the following keyboard functionality to the table:

- Navigate through rows with arrow up/down
- Expand/close Expandable rows with arrow right/left
- Multiselect rows via Shift + Arrow up/down
- Reorder row(s) via Alt + Arrow up/down
- Add a custom Enter functionality to a row (e.g. browser "Apply")

The hook works perfectly with Readonly-Tables, however with editable tables (especially with Tables that have a SelectCell) it does not behave perfectly (see Table --> Edit --> Keyboard Support).
Within Table --> Row / Table --> Tree / BrowserView --> Dialog Browser are example to test it yourself (https://jenkins.ivyteam.io/job/ui-components/job/make-table-keyboard-friendly/11/artifact/packages/components/storybook-static/index.html?path=/)

Something i will do in a later PR are:
- Tests
- (try) Improving hook for Editable-Tables

### DIFF
--- a/packages/components/src/components/common/table/hooks/hooks.tsx
+++ b/packages/components/src/components/common/table/hooks/hooks.tsx
@@ -1,5 +1,5 @@
 import { SearchInput } from '@/components/common/input/input';
-import { selectRow } from '@/utils/table/table';
+import { resetAndSetRowSelection, selectRow } from '@/utils/table/table';
 import {
   getFilteredRowModel,
   getSortedRowModel,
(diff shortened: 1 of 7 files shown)


ivy-lli:
lazy currently also seems not to work (e.g. browserView -> attribute -> requester)

packages/components/src/components/common/table/table.tsx
```diff
@@ -8,7 +8,7 @@ import { Field } from '@/components/common/field/field';
  */
 const Table = React.forwardRef<HTMLTableElement, React.HTMLAttributes<HTMLTableElement>>(({ className, ...props }, ref) => (
   <Field className={cn(root, 'ui-table-root')}>
-    <table ref={ref} className={cn(table, className, 'ui-table')} {...props} />
+    <table ref={ref} className={cn(table, className, 'ui-table')} tabIndex={0} {...props} />
```
ivy-lli, line 11: please add a the :focus-visible to the table (e.g. on firefox you have no idea that the table has focus)

packages/components/src/components/common/table/tree/tree.tsx
```diff
@@ -29,6 +29,7 @@ const expanedButton = <TData,>(row: Row<TData>, lazy?: LazyExpand<TData>) => {
         className={expandButton}
         aria-label={row.getIsExpanded() ? 'Collapse row' : 'Expand row'}
         data-state={row.getIsExpanded() ? 'expanded' : 'collapsed'}
+        tabIndex={-1}
```
ivy-lli, line 32: seems fine for me 👍

ivy-edp:
Yes, you are right. I tried to support lazy, but unfortunately, it wasn’t quite that simple. The "isLoaded" flag and "loadchildren-function" (as I understood it) are specifically tied to browser tables (BrowserNode objects). However, I want this hook to work for all types of "read-only tables." I think I could write a more specific function here that directly expects a table with BrowserNode<TData> objects. This way, I could definitely make it work.

packages/components/src/components/common/table/hooks/hooks.tsx
ivy-lli, line 1: Do you like this code? Because I'm not. 
I'm not the advocate of clean code, but I'm also not a fan of spaghetti code and of duplications
- both methods share many functionalities
- why is the stuff (e.g. what happens on ArrowUp) not in its own function?
- Why are there even two different functions? now the first one only supports single selection but with table expand and the second multi selection with reorder. what e.g. about multiselection without reoder? why do I have to choose? why is this not controlled by options (by the way if the table supports multi selection or not is already there in the table object). 
I personally would expect one function which does everything based on options (e.g. if i want to have reorder stuff then I have to pass the reorder function).
- Now this root variables are calculated everytime the hook rerenders. I cannot imagine that this is very good for the performance... Why is this even needed? we only listen to events and should only calculate stuff needed for the specific actions (I think the only exception is the root selected row, somewhere we need to save that).
- Speeking of the root selected row. I'm pretty sure that the useEffect is not needed if you don't calculate this root variables

ivy-lli (changes requested):

packages/components/src/components/editor/browser/browser.tsx
```diff
@@ -176,7 +177,7 @@ const BrowsersView = ({ browsers, apply, applyBtn }: BrowsersViewProps) => {
                       onChange={browser.globalFilter.setFilter}
                     />
                     <div className={overflowAuto}>
-                      <Table>
+                      <Table onKeyDown={e => browser.handleKeyDown(e, applyHandler)}>
```
ivy-lli, line 180: what is this?

packages/components/src/components/common/table/hooks/hooks.tsx
```diff
@@ -119,3 +119,143 @@ export const useMultiSelectRow = <TData,>(table: Table<TData>) => {
...
+  return { handleKeyDown };
+};
+
+const handleEnterKey = <TData,>(selectedRows: Array<Row<TData>>, onEnterAction?: (row: Row<TData>) => void) => {
+  if (selectedRows.length === 1 && onEnterAction) {
+    onEnterAction(selectedRows[0]);
+  }
+};
+
+const toggleExpand = <TData,>(rows: Row<TData>[], expand: boolean, loadChildren?: (row: Row<TData>) => void) => {
```
ivy-lli, line 229: you only work with the first selected row here, so I think I would represent that in the function signature by only allow to give one row...

packages/components/src/components/common/table/hooks/hooks.tsx
```diff
@@ -119,3 +119,143 @@ export const useMultiSelectRow = <TData,>(table: Table<TData>) => {
...
+  if (rows.length === 1) {
+    if (expand && !rows[0].getIsExpanded()) {
+      if (loadChildren) {
+        loadChildren(rows[0]);
```
ivy-lli, line 233:
```suggestion
      loadChildren.?(rows[0]);
```

packages/components/src/components/common/table/hooks/hooks.tsx
```diff
@@ -119,3 +119,143 @@ export const useMultiSelectRow = <TData,>(table: Table<TData>) => {
...
+      default:
+        break;
+    }
+  };
+
+  const handleArrowKeyUpDown = (event: React.KeyboardEvent<HTMLTableElement>) => {
+    const { multiSelect = false, reorder } = options || {};
+    const allRows = table.getRowModel().rows;
+    const selectedRows = table.getSelectedRowModel().flatRows;
+    const direction = event.key === 'ArrowUp' ? -1 : 1;
```
ivy-lli, line 168: You now check this key twice. I think it is better if you do it like the ArrowRight/Left way. With a direction param

packages/components/src/components/common/table/hooks/hooks.tsx
```diff
@@ -119,3 +119,143 @@ export const useMultiSelectRow = <TData,>(table: Table<TData>) => {
...
+
+    const newReorderIndex = calculateNewReorderIndex(
+      direction,
+      allRows.indexOf(selectedRows[0]),
+      allRows.indexOf(selectedRows[selectedRows.length - 1]),
+      selectedRows.length,
+      allRows.length
+    );
```
ivy-lli, line 176: instead of a function with 5 params, maybe it would be easier if you use an object, then each param has a name

ivy-lli (commented):
Much better 👍 
Still some places for improvements IMO

And you didn't improve the focus-visbile indicator right? Can you do this too pls.

packages/components/src/components/common/table/hooks/hooks.tsx
```diff
@@ -119,3 +119,143 @@ export const useMultiSelectRow = <TData,>(table: Table<TData>) => {
...
+  options?: KeyHandlerOptions<TData>;
+}
+
+export const useTableKeyHandler = <TData,>({ table, data, options }: TableKeyboardHandlerProps<TData>) => {
+  const [rootIndex, setRootIndex] = React.useState<number | undefined>(undefined);
+
+  const handleKeyDown = (event: React.KeyboardEvent<HTMLTableElement>, onEnterAction?: (row: Row<TData>) => void) => {
+    event.stopPropagation();
+
+    switch (event.key) {
```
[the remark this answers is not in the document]
ivy-edp: I agree with the breaks. What would be a good alternative? A Record? Maybe something like this?
grafik

packages/components/src/components/editor/browser/browser.tsx
ivy-edp: The new useTableKeyHandler hook requires the table and data as parameters. However, accessing the data array directly in the BrowserView-Component was challenging, particularly when trying to retrieve a specific browser-table from the browsers array. To simplify this, I directly initialized the handleKeyDown function inside the useBrowser hook and returned it from there. This approach allowed me to then easily set the specific function directly on the <Table> tag.
And also because the applyHandler (which should be run when enter is pressed on a row), is only defined within the BrowserView, so I pass the parameter directly in the function instead of the initial useTableKeyHanlder hook.

packages/components/src/components/common/table/hooks/hooks.tsx
```diff
@@ -119,3 +119,143 @@ export const useMultiSelectRow = <TData,>(table: Table<TData>) => {
...
+      } else {
+        selectedRows[selectedRows.length - 1].toggleSelected();
+      }
+    }
+  };
+
+  return { handleKeyDown };
+};
+
+const handleEnterKey = <TData,>(selectedRows: Array<Row<TData>>, onEnterAction?: (row: Row<TData>) => void) => {
```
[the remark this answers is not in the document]
ivy-edp: Yes, that’s correct. For now, I would also reduce it to a single row since the onEnterAction only works with one row anyway. I would expand it to support multiple rows only once we have a clear use case for it. Additionally, I can do without the handleEnterKey method and directly call the onEnterAction function, as I’m expecting only a single row in any case.
grafik

packages/components/src/components/common/table/hooks/hooks.tsx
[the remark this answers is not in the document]
ivy-lli: Yes why not 👍

packages/components/src/components/common/table/hooks/hooks.tsx
```diff
@@ -119,3 +119,129 @@ export const useMultiSelectRow = <TData,>(table: Table<TData>) => {
...
+
+  const handleArrowKeyUpDown = (event: React.KeyboardEvent<HTMLTableElement>, direction: -1 | 1) => {
+    const { multiSelect = false, reorder } = options || {};
+    const allRows = table.getRowModel().rows;
+    const selectedRows = table.getSelectedRowModel().flatRows;
+
+    const newReorderIndex = calculateNewReorderIndex({
+      direction,
+      firstSelectedRowIndex: allRows.indexOf(selectedRows[0]),
+
```
ivy-lli, line 161: why this new line?

ivy-lli (approved):
Thanks for the improvements. Let it fly ;)

PR #205, opened 2024-12-09
XIVY-15134 treat undefined as empty item in select

ivy-lgi:
In the data class editor, I receive undefined for the cardinality of a field with the initial data if it is not set. In this case, the empty item is shown in the select even though nothing seems to be selected. So let's consider undefined when checking whether the empty item is shown by simply checking if the value is truthy.

empty-item


ivy-lli (approved):
Hopefully this has no sideeffects...

PR #206, opened 2024-12-11
Use import without deconstruction to ensure proper bundling

martin-fleck-at:


ivy-lli (approved):
Thanks! Yes this seems to work. But why? 
Does this make any sense to you or how did you find this solution?
Or is destructuring not supported in lazy imports and library builds?

However the strange url warning in the inscription-client package build still occurs, but it doesn't seem to be the problem then...

ivy-lli:
Hm strange stuff, but I'm glad you found this solution.
I don't think that we use this file, so I think we can live with this warning. I was just curious if you found out something about it.

The other thing that is interesting to me is the update to the latest monaco / monaco language client. I'm not sure if we could benefit there from some new stuff, but I think we need to get better there at the ability to update this it.

I think I merge this for now, but I'm not sure if we should already merge this change back to the product. We are currently still on stabilizing stuff for our current LTS 12. I think I will apply this change then to the next LE version.

PR #207, opened 2024-12-12
Pre work for React 19 support

ivy-lli:
- Fix vulnerabilities
- Update storybook to v8.4.7
- Update testing, lint, etc dev deps
- Update some component libs

There are still transient libraries which does not support react 19 yet (in storybook and radix-ui)


ivy-lmu (approved):
thx 👍

PR #208, opened 2024-12-12
Improve useTableKeyHandler-Hook

ivy-edp:
While I integrated keyboard support into the other editors, I noticed a few bugs/improvement points. I have addressed them here:
- When a scrollbar is present, the selected row is now always visible (centered).
- The Tab key only resets the selection if specifically defined in the options.
- The ArrowLeft/Right functionality was reversed.

### DIFF
--- a/packages/components/src/components/common/table/hooks/hooks.tsx
+++ b/packages/components/src/components/common/table/hooks/hooks.tsx
@@ -124,6 +124,7 @@ interface KeyHandlerOptions<TData> {
   multiSelect?: boolean;
   reorder?: { updateOrder?: (moveIndexes: number[], toIndex: number) => void; getRowId?: (row: TData) => string };
   lazyLoadChildren?: (row: Row<TData>) => void;
+  resetSelectionOnTab?: boolean;
 }
 
 interface TableKeyboardHandlerProps<TData> {
@@ -141,9 +142,9 @@ export const useTableKeyHandler = <TData,>({ table, data, options }: TableKeyboa
     const actions: Record<string, () => void> = {
       ArrowUp: () => handleArrowKeyUpDown(event, -1),
       ArrowDown: () => handleArrowKeyUpDown(event, 1),
-      ArrowLeft: () => toggleExpand(table.getSelectedRowModel().flatRows[0], true, options?.lazyLoadChildren),
-      ArrowRight: () => toggleExpand(table.getSelectedRowModel().flatRows[0], false, options?.lazyLoadChildren),
-      Tab: () => table.resetRowSelection(),
+      ArrowLeft: () => toggleExpand(false, table.getSelectedRowModel().flatRows[0], options?.lazyLoadChildren),
+      ArrowRight: () => toggleExpand(true, table.getSelectedRowModel().flatRows[0], options?.lazyLoadChildren),
+      Tab: () => options?.resetSelectionOnTab && table.resetRowSelection(),
       Enter: () => onEnterAction?.(table.getSelectedRowModel().flatRows[0])
     };
 
@@ -151,6 +152,7 @@ export const useTableKeyHandler = <TData,>({ table, data, options }: TableKeyboa
   };
 
   const handleArrowKeyUpDown = (event: React.KeyboardEvent<HTMLTableElement>, direction: -1 | 1) => {
+    event.preventDefault();
     const { multiSelect = false, reorder } = options || {};
     const allRows = table.getRowModel().rows;
     const selectedRows = table.getSelectedRowModel().flatRows;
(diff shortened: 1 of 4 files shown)


packages/components/src/components/common/table/hooks/hooks.tsx
```diff
@@ -175,6 +177,7 @@ export const useTableKeyHandler = <TData,>({ table, data, options }: TableKeyboa
       table.resetRowSelection();
       allRows[newReorderIndex].toggleSelected();
       setRootIndex(newReorderIndex);
+      event.currentTarget.rows[newReorderIndex].scrollIntoView({ block: 'center' });
```
ivy-lli, line 180: no negative effect if no scroll bar?

packages/components/src/components/common/table/hooks/hooks.tsx
```diff
@@ -209,10 +212,10 @@ export const useTableKeyHandler = <TData,>({ table, data, options }: TableKeyboa
 };
 
 const toggleExpand = <TData,>(row: Row<TData>, expand: boolean, loadChildren?: (row: Row<TData>) => void) => {
-  if (expand && !row.getIsExpanded()) {
+  if (expand && row.getIsExpanded && !row.getIsExpanded()) {
```
ivy-lli, line 215: why needed?

packages/components/src/components/common/table/hooks/hooks.tsx
```diff
@@ -209,10 +212,10 @@ export const useTableKeyHandler = <TData,>({ table, data, options }: TableKeyboa
...
     loadChildren?.(row);
     row.toggleExpanded();
-  } else if (!expand && row.getIsExpanded()) {
+  } else if (!expand && row.getIsExpanded && row.getIsExpanded()) {
```
ivy-lli, line 218: same why needed?

packages/components/src/components/common/table/hooks/hooks.tsx
ivy-edp: I have tested it a little and I have not noticed any. If you like, you are welcome to try it out a little and check if you notice something, but I don't think there should be any negative effects, as the arrows within a table don't have any effect on a table without a scrollbar.

packages/components/src/components/common/table/hooks/hooks.tsx
ivy-edp: I had some strange getIsExpanded is undefined errors in the browser. But I couldn't reproduce them today. However, I just got an "row is undefined" error when trying to use the arrowleft/right button while no row is selected. I will now also check if the row is not undefined and can be expanded.

ivy-lli: the point is that typescript should make sure that the row is not undefined.
So either we should mark the row here as optional, so typescript can help us or identify the real problem and check it there.
To be clear, I think this is the array get syntax which is not clear that it can also return undefined. So I think we should mark the row here as undefined and check it.

packages/components/src/components/common/table/hooks/hooks.tsx
```diff
@@ -209,10 +212,10 @@ export const useTableKeyHandler = <TData,>({ table, data, options }: TableKeyboa
 };
 
 const toggleExpand = <TData,>(row: Row<TData>, expand: boolean, loadChildren?: (row: Row<TData>) => void) => {
-  if (expand && !row.getIsExpanded()) {
+  if (expand && row && row.getCanExpand() && !row.getIsExpanded()) {
```
ivy-lli, line 215: mark the row as optional so it is clear
```suggestion
const toggleExpand = <TData,>(expand: boolean, row?: Row<TData>, loadChildren?: (row: Row<TData>) => void) => {
  if (expand && !row.getIsExpanded()) {
  if (expand && row && row.getCanExpand() && !row.getIsExpanded()) {
```

packages/components/src/components/common/table/hooks/hooks.tsx
ivy-edp: Not 100% sure if i understood correctly... I marked the row as optional now and check at the start, if the row is undefined. Something like this:
grafik

ivy-lli: When we don't mark the row as optional here then what happens? We lie to ourselfes and don't help anyone which uses this code.

Typescript is only a layer on top of JS, for js it don't makes any difference if this variable is define or not, but in the editor we want to have the best experience. and we don't get there when we lie to ourselves. This row can be undefined (even if TypeScript is not clever enough to recognize that), so we should help ourself by mark it like that so then typescript can help us catch errors.

ivy-lli (approved):

PR #218, opened 2025-01-08
XIVY-15631 Add react-hotkeys-hook, with hotkey helper text function

ivy-lli:
- Deprecate useShortcut hook
- Add react-hotkeys-hook lib
- Add hotkeyText utility function

### DIFF
--- a/packages/components/src/hooks/hotkey.stories.tsx
+++ b/packages/components/src/hooks/hotkey.stories.tsx
@@ -1,0 +1,56 @@
+import { hotkeyText } from '@/utils/hotkey';
+import type { Meta, StoryObj } from '@storybook/react';
+import { useState } from 'react';
+import { useHotkeys } from 'react-hotkeys-hook';
+
+const meta: Meta = {
+  title: 'Hooks/Hotkey',
+  tags: ['autodocs']
+};
+
+export default meta;
+
+export const Default: StoryObj = {
+  render: () => {
+    const [count, setCount] = useState(0);
+    useHotkeys('mod+alt+n', () => setCount(prevCount => prevCount + 1));
+    const hotkey = hotkeyText('mod+alt+n');
+
+    return (
+      <span>
+        Pressed the <b>{hotkey}</b> key {count} times.
+      </span>
+    );
+  }
+};
+
+export const Scoped: StoryObj = {
+  render: () => {
+    const [count, setCount] = useState(0);
+    const ref = useHotkeys('c', () => setCount(prevCount => prevCount + 1));
+    return (
+      <div ref={ref} tabIndex={-1} style={{ border: '2px solid #9e768f' }}>
+        <h1>Scoped Hotkeys</h1>
+        <p>Focus me, and press c to increment the count</p>
+        <p>Count: {count}</p>
+      </div>
+    );
+  }
+};
+
+export const Global: StoryObj = {
+  render: () => {
+    const [count, setCount] = useState(0);
+    useHotkeys('q', () => setCount(prevCount => prevCount + 1));
+    useHotkeys('w', () => setCount(prevCount => prevCount - 1));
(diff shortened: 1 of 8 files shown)


ivy-lli:
The existing solution was not bad, so why replace it?:
- Doesn't work correctly (at least om mac) for e.g. shortcuts like `Cmd+Alt+a` returns on event.key = "å". So it is the modified key which is not helpful for such detections. The alternative is to use event.code that returns always `KeyA` but this makes checking complicater
- The hook lacks at least one big functionality, register hotkeys only for a specific scope (not whole document)

I think it makes more sense to use a existing battle teseted solution.

packages/components/src/utils/hotkey.ts
```diff
@@ -0,0 +1,10 @@
+const isMac = () => window.navigator.userAgent.indexOf('Mac') !== -1;
+
+type ModKey = '⌘' | 'CTRL';
+
+const modKey = (): ModKey => (isMac() ? '⌘' : 'CTRL');
+
+export const MOD_KEY = modKey();
+export const IS_MAC = isMac();
+
+export const hotkeyText = (hotkey: string): string => hotkey.replace('mod', modKey()).replace('shift', '⇧');
```
ivy-lgi, line 10: Not sure if low-coders understand the symbol `⇧` as SHIFT. Is this common?

packages/components/src/hooks/hotkey.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+  expect(hotkeyText('ctrl+a')).toEqual('ctrl+a');
+  expect(hotkeyText('shift+a')).toEqual('⇧+a');
+  expect(hotkeyText('alt+a')).toEqual('alt+a');
+  expect(hotkeyText('options+a')).toEqual('options+a');
+  expect(hotkeyText('mod+a')).toEqual('CTRL+a');
+
+  vi.stubGlobal('navigator', { userAgent: 'Mac' });
+  expect(hotkeyText('mod+a')).toEqual('⌘+a');
+  expect(hotkeyText('mod+shift+a')).toEqual('⌘+⇧+a');
+});
```
ivy-lgi, line 21: Duplicated test? It's already present in `packages/components/src/utils/hotkey.test.tsx`

packages/components/src/hooks/hotkey.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+  title: 'Hooks/Hotkey',
+  tags: ['autodocs']
+};
+
+export default meta;
+
+export const Default: StoryObj = {
+  render: () => {
+    const [count, setCount] = useState(0);
+    useHotkeys('mod+shift+d', () => setCount(prevCount => prevCount + 1));
```
ivy-lgi, line 16: Maybe not the best example to use as `mod+shift+d` is already taken in firefox at least (New Bookmarks). I suggest `mod+alt+d`.

packages/components/src/utils/hotkey.ts
ivy-lli: I have a ⇧ on my keyboard not a SHIFT ;)

packages/components/src/hooks/hotkey.test.tsx
ivy-lli: ah copy paste failure thx

packages/components/src/hooks/hotkey.stories.tsx
ivy-lli: hm tested this on my machine there this worked but yes, will check it again and adjust 👍

ivy-lgi: It does work but it also opens a modal window at the same time
image

ivy-lgi (approved):

PR #234, opened 2025-01-16
XIVY-13545 Add poss. to configure delete-key behaviour on row table

ivy-edp:
With this change, I am adding functionality to control not only the Enter key but also the Delete key, allowing you to define what happens when the Delete key is clicked on a row. Additionally, I am introducing a new key object, which will make it much easier to extend in the future if more key functionalities need to be enabled for a row, without having to modify every place where this hook is used.

### DIFF
--- a/packages/components/src/components/editor/browser/browser.tsx
+++ b/packages/components/src/components/editor/browser/browser.tsx
@@ -177,7 +177,7 @@ const BrowsersView = ({ browsers, apply, applyBtn }: BrowsersViewProps) => {
                       onChange={browser.globalFilter.setFilter}
                     />
                     <div className={overflowAuto}>
-                      <Table onKeyDown={e => browser.handleKeyDown(e, applyHandler)}>
+                      <Table onKeyDown={e => browser.handleKeyDown(e, { onEnterAction: applyHandler })}>
                         <TableBody>
                           {browser.table.getRowModel().rows?.length ? (
                             browser.table.getRowModel().rows.map(row => (
--- a/packages/components/src/components/common/table/hooks/hooks.tsx
+++ b/packages/components/src/components/common/table/hooks/hooks.tsx
@@ -146,6 +146,11 @@ interface KeyHandlerOptions<TData> {
   resetSelectionOnTab?: boolean;
 }
 
+export interface KeyDownOptions<TData> {
+  onEnterAction?: (row: Row<TData>) => void;
+  onDeleteAction?: (row: Row<TData>) => void;
+}
+
 interface TableKeyboardHandlerProps<TData> {
   table: Table<TData>;
   data: Array<TData>;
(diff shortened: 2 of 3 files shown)


ivy-lli (approved):
Ok for me, but I've already implemented the delete shortcut for config and data class editor.
So I'm fine with add this to this hook too, but then pls improve the event.stopPropagation code, so it only blocks the propagation if there is a onDeleteAction (best also same for the others e.g. enter) defined. Otherwise you will break stuff.

ivy-edp:
> Ok for me, but I've already implemented the delete shortcut for config and data class editor. So I'm fine with add this to this hook too, but then pls improve the event.stopPropagation code, so it only blocks the propagation if there is a onDeleteAction (best also same for the others e.g. enter) defined. Otherwise you will break stuff.

Ah, I see. I wasn't aware that you had implemented delete and combine as well. I think I will remove this PR again, as I think your solution is a bit more flexible (you don't always have to touch the UI components first if a new shortcut needs to be integrated). Also, if I would merge this change, I had to do quiet some changes on multiple repos (process editor/dataclass editor/variable editor in both the master and release12 branches). Or do you see a major benefit if we would move the table shortcut handling (delete and c) into this hook?

ivy-lli:
> > Ok for me, but I've already implemented the delete shortcut for config and data class editor. So I'm fine with add this to this hook too, but then pls improve the event.stopPropagation code, so it only blocks the propagation if there is a onDeleteAction (best also same for the others e.g. enter) defined. Otherwise you will break stuff.
> 
> Ah, I see. I wasn't aware that you had implemented delete and combine as well. I think I will remove this PR again, as I think your solution is a bit more flexible (you don't always have to touch the UI components first if a new shortcut needs to be integrated). Also, if I would merge this change, I had to do quiet some changes on multiple repos (process editor/dataclass editor/variable editor in both the master and release12 branches). Or @ivy-lli do you see a major benefit if we would move the table shortcut handling (delete and c) into this hook?

Not really. I think it is cool to have the whole select stuff here, but the others are more or less only redirect things. Not bad, but maybe sometimes not what you really want.
I'm not against this change or addition, and if you make the changes according the stopPropagation as I said I also don't think that you have to change anything else in another repo.

packages/components/src/components/editor/browser/browser.tsx
```diff
@@ -177,7 +177,7 @@ const BrowsersView = ({ browsers, apply, applyBtn }: BrowsersViewProps) => {
                       onChange={browser.globalFilter.setFilter}
                     />
                     <div className={overflowAuto}>
-                      <Table onKeyDown={e => browser.handleKeyDown(e, applyHandler)}>
+                      <Table onKeyDown={e => browser.handleKeyDown(e, { onEnterAction: applyHandler })}>
```
ivy-edp, line 180: @ivy-lli Unfortunately, I would need to make the following adjustments to all repositories where I've used the hook and defined an enter behavior, as they now require an options object. However, I've decided not to make this change and will keep everything as it currently stands. If a use case arises in the future where this would be beneficial, it can be quickly addressed then.

ivy-lli: Ah yes. Good for me 👍

PR #246, opened 2025-01-24
Improve condition builder

ivy-lli:
- Style problems (do not use a field, if you don't use label or something else)
- Use basic input instead of input
- Do not leak provider outside of component
- Restructure and rename components


ivy-lli:
This causes changes e.g. like:
image

ivy-edp (approved):
Cool, thanks. Looks good to me 👍 
I guess with "Do not leak provider outside of component" you mean that now you can directly us the ConditionBuilder-Component and within this Component the Provider is set without the need of wrapping the ConditionBuilder-Compontent first around the Provider.

PR #250, opened 2025-01-29
XIVY-15849 make websocket abortable

ivy-lmu:


packages/jsonrpc/src/ws-connection-provider.ts
```diff
@@ -49,7 +49,11 @@ const defaultOptions: Required<WebSocketOptions> = {
   closeMessage: 'Connection was closed, will not reconnect!'
 } as const;
 
-export const webSocketConnection = <TConnection = Connection>(url: string | URL, initOptions?: WebSocketOptions) => {
+export const webSocketConnection = <TConnection = Connection>(
+  url: string | URL,
+  initOptions?: WebSocketOptions,
+  abortSignal?: AbortSignal
```
ivy-lli, line 55: I think I would move this to the options (or initOptions).

ivy-lli (approved):
Yes imagined something like that, this works then?

packages/jsonrpc/src/ws-connection-provider.ts
ivy-lli: otherwise you will need to pass something empty if you want to add this abortSignal
`webSocketConnection('url', undefined, abortSignal)`
or
`webSocketConnection('url', {abortSignal})`

ivy-lmu:
> Yes imagined something like that, this works then?

yup, see prototype -> https://github.com/axonivy/neo/compare/master...close-ws2

PR #253, opened 2025-01-30
Border focused editcell

ivy-edp:
I had difficulty seeing which cell was focussed (especially with SelectCell). I have now added the focus-border. Unfortunately I couldn't reuse vars.borders.active here, because for some reason the active border is set to the blue backgroundcolour of the row when rows are selected and so you can't see the border.

borderforfocusededitcell

### DIFF
--- a/packages/components/src/components/common/table/edit/edit.test.tsx
+++ b/packages/components/src/components/common/table/edit/edit.test.tsx
@@ -43,3 +43,92 @@ test('combobox', async () => {
   expect(screen.getAllByRole('row')[1]).toHaveAttribute('data-state', 'selected');
   expect(screen.getAllByRole('combobox')[1]).toHaveValue('123');
 });
+
+const setupTable = async () => {
+  render(<Table />);
+  const rows = screen.getAllByRole('row');
+  const user = userEvent.setup();
+  expect(rows[1]).toHaveAttribute('data-state', 'unselected');
+  await user.click(rows[1]);
+  expect(rows[1]).toHaveAttribute('data-state', 'selected');
+  expect(rows[2]).toHaveAttribute('data-state', 'unselected');
+  return { rows, user };
+};
+
+const selectRowAndNavigate = async (rows: HTMLElement[], user: ReturnType<typeof userEvent.setup>, colIndex: number) => {
+  // Navigate to the correct column
+  for (let i = 0; i <= colIndex; i++) {
+    await user.keyboard('[Tab]');
+  }
+  const tagName = colIndex === 0 ? 'button' : 'input';
+  expect(rows[1].getElementsByTagName('td')[colIndex].getElementsByTagName(tagName)[0]).toHaveFocus();
+
+  // Move down two rows
+  await user.keyboard('[ArrowDown]');
+  await user.keyboard('[ArrowDown]');
+
+  expect(rows[1]).toHaveAttribute('data-state', 'unselected');
+  expect(rows[3]).toHaveAttribute('data-state', 'selected');
+  expect(rows[3].getElementsByTagName('td')[colIndex].getElementsByTagName(tagName)[0]).toHaveFocus();
+
+  // Check jump over disabled select cell
+  await user.keyboard('[ArrowDown]');
+  await user.keyboard('[ArrowDown]');
+  await user.keyboard('[ArrowDown]');
+
+  expect(rows[8]).toHaveAttribute('data-state', 'selected');
+  expect(rows[3]).toHaveAttribute('data-state', 'unselected');
+  expect(rows[8].getElementsByTagName('td')[colIndex].getElementsByTagName(tagName)[0]).toHaveFocus();
+
+  await user.keyboard('[ArrowUp]');
+  expect(rows[6]).toHaveAttribute('data-state', 'selected');
+  expect(rows[8]).toHaveAttribute('data-state', 'unselected');
+  expect(rows[6].getElementsByTagName('td')[colIndex].getElementsByTagName(tagName)[0]).toHaveFocus();
+};
+
+test('keyboard navigation with arrow for select cell', async () => {
+  const { rows, user } = await setupTable();
+  await selectRowAndNavigate(rows, user, 0);
+});
+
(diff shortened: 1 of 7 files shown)


packages/components/src/components/common/table/table.css.ts
```diff
@@ -60,6 +60,9 @@ export const cell = style({
   selectors: {
     [`&:has(.ui-table-edit-cell)`]: {
       padding: 0
+    },
+    [`&:has(.ui-table-edit-cell:focus), &:has(.ui-table-edit-cell:focus-within)`]: {
+      border: `1px solid black`
```
ivy-lli, line 65: Big chance that this will not work great in dark theme...
Does the focus-shadow not work?
`boxShadow: vars.shadow.focus`

ivy-lli: And this focus will be visible always no only when you use your keyboard...

But maybe this is indeed good, but maybe just change the background a bit would also look good

ivy-edp: no this does unfortunatly not work. But yes in does not look that pleasent in the dark theme. i will have a closer look...

ivy-edp: Acctually it does not look that bad in dark theme:
grafik

Also we can not set the border to white as the selected row already is white and than you cant see the focus-border.

ivy-lli (approved):

ivy-edp:
I was able to accieve the functionality we were looking for. However I think I still can improve the selectNextPreviousCell-function. I will have a closer look on it tomorrow.
You are now able to open the select and combobox menu only by using ENTER. And as soon as the menu is open you can navigate using the arrowkeys
edittableKeyboardsupport

packages/components/src/components/common/table/edit/edit.tsx
```diff
@@ -53,8 +94,59 @@ type ComboCellProps<TData, TCombo extends ComboboxOption> = Omit<ComboboxProps<T
...
 
 export { InputCell, SelectCell, ComboCell };
+
+const selectNextPreviousCell = (htmlElement: HTMLButtonElement | HTMLInputElement, direction: -1 | 1) => {
+  const focusedCell = htmlElement.closest('td');
+  if (focusedCell && focusedCell.tagName.toLowerCase() === 'td') {
+    const selectedRow = focusedCell.parentElement;
+    if (selectedRow && selectedRow.tagName.toLowerCase() === 'tr') {
+      const cellIndex = Array.from(focusedCell.parentElement?.children || []).indexOf(focusedCell);
+      const rowIndex = Array.from(selectedRow.parentElement?.children || []).indexOf(selectedRow);
```
ivy-lli, line 128: can't you get this stuff from the table cell?

packages/components/src/components/common/table/table.css.ts
ivy-lli, line 1: do you have wrong line endings that git is not able to show a good diff here?

ivy-lli (commented):
Cool, but we defenitly need tests for this!

packages/components/src/components/common/table/edit/edit.tsx
ivy-edp: Yes, you are right. To get the cell- and rowindex, I could use the cell from the tanstack table. But to set the focus directly in the input/button, I still need the HTMLElement. Or do you know if there is a supported way from tanstack to focus a cell? I quickly had a look but couldn't find a way to focus a cell.

packages/components/src/components/common/table/table.css.ts
ivy-edp: yes sorry, should now have the correct line endings

ivy-edp:
I now have:
- Added tests
- for Cell- & RowIndex I use the cell
- onKeyDown added to the TableAddRow (I need it later to return the focus to the table after adding)
- correct line ending

packages/components/src/components/common/table/footer/footer.tsx
```diff
@@ -16,6 +25,7 @@ const TableAddRow = ({ addRow, tabindex }: { addRow: () => void; tabindex?: numb
         variant='outline'
         aria-label='Add row'
         tabIndex={tabindex}
+        onKeyDown={onKeyDown}
```
ivy-lli, line 28: why needed?

ivy-lli: you could also imply allow all possible button props instead of add all one by one...

packages/components/src/components/common/table/edit/edit.test.tsx
```diff
@@ -43,3 +43,87 @@ test('combobox', async () => {
...
+  const rows = screen.getAllByRole('row');
+  const user = userEvent.setup();
+  expect(rows[1]).toHaveAttribute('data-state', 'unselected');
+  await user.click(rows[1]);
+  expect(rows[1]).toHaveAttribute('data-state', 'selected');
+  expect(rows[2]).toHaveAttribute('data-state', 'unselected');
+  return { rows, user };
+};
+
+const selectRowAndNavigate = async (rows: HTMLElement[], user: ReturnType<typeof userEvent.setup>, colIndex: number) => {
```
ivy-lli, line 58: you never test ArrowUp?

ivy-lli (approved):
Good some comments below:

packages/components/src/components/common/table/edit/edit.tsx
```diff
@@ -119,17 +119,20 @@ ComboCell.displayName = 'ComboCell';
 
 export { InputCell, SelectCell, ComboCell };
 
-const selectNextPreviousCell = <TData,>(
-  htmlElement: HTMLButtonElement | HTMLInputElement,
+export const selectNextPreviousCell = <TData,>(
```
ivy-edp, line 122: @ivy-lli is the export enought so that i can access this function within the process-editor-client?

ivy-lli: if the root index in the components/src exports all of this file then yes. But this should be the case otherwise we could also not use the InputCell etc.

PR #336, opened 2025-02-14
Add graph component

ivy-edp:


ivy-edp:
Hi @ivy-lli 
I’ve created an initial version of the graph component. It would be great if you could take a look, play around with the example graph, and provide feedback.

Here’s an overview of how it works:
- The component is designed for easy use: simply provide an array of graphNodes and graphEdges.
- Data needs to be prepared beforehand:
 - CustomNodeData for each node
 - EdgeData for the relationships between nodes
- Filtering is possible:
 - Double-click on a node or use the select menu to only show connected nodes.
- I’ve added two buttons at the top-right to reorder the graph using the Dagre algorithm (initial reorder is already applied).
- Features like the minimap and controls are active, though they could be made configurable in the options.

It’d be great if you could test this out. We can discuss it further on Monday, either via phone or text, whichever you prefer.

One thing I’m still unsure about is how easily we can transfer the data we receive from the backend into CustomNodeData and NodeEdges.

ivy-edp:
@ivy-lli I’ve completed and improved the graph component. Could you take a look? We can also do a desk review if you prefer.
**General**
- I am using React Flow to render the graph.
- I am using the Dagre algorithm to automatically layout the graph.
- I have tried to minimize the use of useEffect and, whenever possible, have outsourced logic to useCallback, custom hooks, or standalone functions.

**Features**
- The component can render a graph by passing an array of CustomNodeData.
  - The CustomNodeData object allows defining various options.
- Users can filter nodes via a SelectMenu or by double-clicking on a node, showing only directly connected nodes while reapplying the layout automatically.
- Nodes can contain custom content that can be expanded or collapsed.
- Nodes can be freely moved around on the canvas.

At some point, I should probably add tests, especially for the useCustomGraph hook.

packages/jsonrpc/package.json
```diff
@@ -21,7 +21,7 @@
   "main": "lib/jsonrpc.js",
   "module": "lib/jsonrpc.js",
   "scripts": {
-    "clean": "rimraf lib *.tsbuildinfo",
+    "clean": "rimraf --glob lib *.tsbuildinfo",
```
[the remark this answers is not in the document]
ivy-lli: hm strange, but should be fine I think

package.json
```diff
@@ -12,7 +12,7 @@
     "generate": "npm run generate --workspace=@axonivy/ui-icons",
     "build": "lerna run build",
     "package": "lerna run package",
-    "clean": "lerna run clean",
+    "clean": "lerna run clean -- --glob",
```
ivy-lli, line 15: why do you need to add it here?

packages/components/src/components/common/select/select.tsx
```diff
@@ -147,7 +149,7 @@ const BasicSelect = ({
       <SelectTrigger className={className} onKeyDown={onKeyDown}>
         <SelectValue placeholder={placeholder} />
       </SelectTrigger>
-      <SelectContent>
+      <SelectContent style={{ width: menuWidth }}>
```
ivy-lli, line 152: why needed?

packages/components/src/components/editor/graph/graph.tsx
```diff
@@ -0,0 +1,119 @@
...
+    expandContent?: boolean;
+  };
+};
+
+export type EdgeData = {
+  id: string;
+  label?: string;
+};
+
+export type GraphNode = Node<{ CustomNodeData: CustomNodeData }, 'custom'>;
```
ivy-lli, line 32: why is this object key uppercase?
```suggestion
export type GraphNode = Node<{ customNodeData: CustomNodeData }, 'custom'>;
```

packages/components/src/utils/graph.ts
ivy-lli, line 1: why is this stuff in utils?
Utils are thing which is also exported and is used by code outside of this components here.
This code here seems only to be used by the edges.

packages/components/src/utils/graph.ts
```diff
@@ -0,0 +1,150 @@
...
+  const [x, y] = getHandleCoordsByPosition(nodeA, position);
+  return [x, y, position];
+}
+
+function getHandleCoordsByPosition(node: InternalNode, handlePosition: Position): [number, number] {
+  const handle = node.internals.handleBounds?.source?.find(h => h.position === handlePosition);
+
+  if (!handle) {
+    throw new Error(`Handle not found for position: ${handlePosition}`);
+  }
```
ivy-lli, line 37: can be dangerous if you throw an error without catch it.
Mostly it is much better to just return undefined, because this will be shown by the typescript compiler so you are enforced to handle this case.
If you throw an error in a function no one will tell the user of this function that there should by a try catch around it.

packages/components/src/utils/graph.ts
ivy-lli: seems that this code should be tested...

packages/components/src/components/editor/graph/nodes/customNode.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const disableHandles = data.CustomNodeData.options?.disableHandles;
+  const [expanded, setExpanded] = useState(false);
+
+  return (
+    <>
+      <Handle type='source' position={Position.Top} id='t' style={{ pointerEvents: 'none', opacity: disableHandles ? 0 : 1 }} />
+      <Handle type='source' position={Position.Left} id='l' style={{ pointerEvents: 'none', opacity: disableHandles ? 0 : 1 }} />
+      <Handle type='source' position={Position.Right} id='r' style={{ pointerEvents: 'none', opacity: disableHandles ? 0 : 1 }} />
+      <Handle type='source' position={Position.Bottom} id='b' style={{ pointerEvents: 'none', opacity: disableHandles ? 0 : 1 }} />
+      <div className={`${customNode} ${selected ? selectedNode : data.CustomNodeData.options?.highlightNode ? originalNode : normalNode}`}>
```
ivy-lli, line 30: maybe use the cn utility

packages/components/src/components/editor/graph/nodes/customNode.tsx
```diff
@@ -0,0 +1,61 @@
...
+              {data.CustomNodeData.info && expanded && <div className={customNodeHeaderInfo}>{data.CustomNodeData.info}</div>}
+            </Flex>
+            <Flex direction='row' gap={1} alignItems='center' style={{ height: '20px' }}>
+              {data.CustomNodeData.options?.controls}
+              {data.CustomNodeData.options?.expandContent && (
+                <>
+                  <Separator decorative orientation='vertical' style={{ marginInline: 'var(--size-1)' }} />
+                  <IvyIcon
+                    icon={IvyIcons.Chevron}
+                    className={`${expander} ${expanded ? expanderOpen : expanderClose}`}
```
ivy-lli, line 44: maybe use cn utility

packages/components/src/components/editor/graph/nodes/customNode.tsx
ivy-lli, line 1: not sure about your usages of style here if this things should not live in the CSS file. But definilty instead of use var(--size-1) or var(--N900) things use the typed variables!

packages/components/src/components/editor/graph/graph.stories.tsx
```diff
@@ -0,0 +1,54 @@
...
+        label: dataClass.name,
+        info: dataClass.fqName.long,
+        content: <CustomNodeFieldContent fields={dataClass.fields} />,
+        target: dataClass.relations,
+        options: {
+          controls: <Button icon={IvyIcons.DataClass} onClick={() => console.log('Open ' + dataClass.name)} />,
+          expandContent: true
+        }
+      };
+    });
```
ivy-lli, line 30: you can do this ouside of the story.

packages/components/src/components/editor/graph/edges/floatingEdge.tsx
```diff
@@ -0,0 +1,49 @@
...
+          <div
+            style={{
+              position: 'absolute',
+              transform: `translate(-50%, -50%) translate(${labelX}px, ${labelY}px)`,
+              background: 'white',
+              padding: '2px 5px',
+              borderRadius: '4px',
+              fontSize: '12px',
+              whiteSpace: 'nowrap'
+            }}
```
ivy-lli, line 39: not sure if this styles should not live in a CSS file. except the coded translate of course.

packages/components/src/components/editor/graph/edges/floatingEdge.tsx
```diff
@@ -0,0 +1,49 @@
...
+    sourceY: sy,
+    sourcePosition: sourcePos,
+    targetPosition: targetPos,
+    targetX: tx,
+    targetY: ty
+  });
+
+  return (
+    <>
+      <path id={id} className='react-flow__edge-path' d={edgePath} strokeWidth={2} markerEnd={markerEnd} style={style} />
```
ivy-lli, line 26: In the ui-components I define the prefixed the "named classes" always with "ui-" and without underlines, e.g. like 'ui-flow-edge-path'

packages/components/src/components/editor/graph/edges/floatingEdge.tsx
```diff
@@ -0,0 +1,49 @@
+import { getEdgeParams } from '@/utils/graph';
+import type { EdgeProps } from '@xyflow/react';
+import { EdgeLabelRenderer, getBezierPath, useInternalNode } from '@xyflow/react';
+
+function FloatingEdge({ id, source, target, markerEnd, style, label }: EdgeProps) {
```
ivy-lli, line 5: sure makes no difference, but I like consistency in the code.
Currently in this codebase components are written with arrow functions (const). you also write most of the code with it, not sure why you mix things up :)

packages/components/src/components/editor/graph/edges/floatingEdge.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const { sx, sy, tx, ty, sourcePos, targetPos } = getEdgeParams(sourceNode, targetNode);
+
+  const [edgePath, labelX, labelY] = getBezierPath({
+    sourceX: sx,
+    sourceY: sy,
+    sourcePosition: sourcePos,
+    targetPosition: targetPos,
+    targetX: tx,
+    targetY: ty
+  });
```
ivy-lli, line 22: Not sure if all of this should be directly in a custom hook...

packages/components/src/components/editor/graph/edges/circleFloatingEdge.tsx
```diff
@@ -0,0 +1,49 @@
+import { getCircleEdgeParams } from '@/utils/graph';
+import type { EdgeProps } from '@xyflow/react';
+import { EdgeLabelRenderer, getBezierPath, useInternalNode } from '@xyflow/react';
+
+function CircleFloatingEdge({ id, source, target, markerEnd, style, label }: EdgeProps) {
```
ivy-lli, line 5: same comments as on the floatingEdge

packages/components/src/components/editor/graph/edges/FloatingConnectionLine.tsx
```diff
@@ -0,0 +1,38 @@
...
+    sourcePosition: fromPosition,
+    targetPosition: toPosition,
+    targetX: toX,
+    targetY: toY
+  });
+
+  return (
+    <g>
+      <path fill='none' stroke='#222' strokeWidth={1.5} className='animated' d={edgePath} />
+      <circle cx={toX} cy={toY} fill='#fff' r={3} stroke='#222' strokeWidth={1.5} />
```
ivy-lli, line 33: you should reuse colors from the theme.

packages/components/src/components/editor/graph/edges/FloatingConnectionLine.tsx
```diff
@@ -0,0 +1,38 @@
...
+    measured: {
+      width: 1,
+      height: 1
+    },
+    internals: {
+      positionAbsolute: { x: toX, y: toY }
+    }
+  };
+
+  const { sx, sy } = getCircleEdgeParams(fromNode, targetNode as InternalNode);
```
ivy-lli, line 20: instead of hard cast here and trick typescript, better define the type at the root and even help typescript define this object for you `const targetNode: InternalNode = {};`

packages/components/src/components/editor/graph/edges/FloatingConnectionLine.tsx
```diff
@@ -0,0 +1,38 @@
+import { getCircleEdgeParams } from '@/utils/graph';
+import { getBezierPath, type ConnectionLineComponentProps, type InternalNode } from '@xyflow/react';
+
+function FloatingConnectionLine({ toX, toY, fromPosition, toPosition, fromNode }: ConnectionLineComponentProps) {
```
ivy-lli, line 4: function / const

packages/components/src/components/editor/graph/data/useCustomGraph.ts
```diff
@@ -0,0 +1,154 @@
...
+  useEffect(() => {
+    setSelectedNode('all');
+    setNodes(getLayoutedElements({ ...mapNodesAndEdges(graphNodes, getNodes(), options), direction: 'TB' }).nodes);
+    setEdges(getLayoutedElements({ ...mapNodesAndEdges(graphNodes, getNodes(), options), direction: 'TB' }).edges);
+  }, [graphNodes, options, getNodes]);
+
+  return { nodes, edges, selectedNode, onNodesChange, onEdgesChange, onConnect, onLayout, onFilterApply };
+};
+
+export default useCustomGraph;
```
ivy-lli, line 83: I think I do not use anywhere default exports in this repo, I would directly export the hook, so it is named.

ivy-lli (commented):
Good start 👍 Some comments however.

In general you mix methods with function and arrow functions, I like consistency as it makes reading code easier IMO. Currently we use more arrow functions so I think we should use them.

Also make sure that you use the theme variables via the typed variables. maybe move more style stuff into CSS files.

And I'm not sure about the therm "Custom", what is custom here? Normally I follow here the concept of modular things, so to have a basic implementation which should be fine for most use cases but with the option to compose the modules on there own if wanted (e.g. like on the Select component). Not sure if this is true in this case, but I do not really like the Custom term as it is not custom here, maybe it would be custom in the code which uses this thing or not?
Maybe you should better follow the other terms, like Root etc: GraphRoot,...

package.json
ivy-edp: You are right for this case the --glob is not needed 👍 For the other cases it is needed to get rid of this rimraf error:
grafik

ivy-lli:
@ivy-edp As this is a big thing (or at least it looks like it), I think it makes sense to have it in its own project.
As I moved storybook to the root, you can now have also stories in other projects. So it should now be possible to have this in it's own package.

ivy-edp:
> @ivy-edp As this is a big thing (or at least it looks like it), I think it makes sense to have it in its own project. As I moved storybook to the root, you can now have also stories in other projects. So it should now be possible to have this in it's own package.

Yes, i can do that 👍 I will add a new package called "package/graph"

packages/components/src/components/editor/graph/edges/floatingEdge.tsx
ivy-edp: Thanks, however this is not a style i defined. This is a style from react-flow itself that is needed, so that the arrows look like this:
grafik

otherwise they would look like this:
grafik

packages/components/src/components/editor/graph/edges/floatingEdge.tsx
ivy-edp: Thanks, these functions that slipped in where due to a copy paste error from the react flow examples ✌️

packages/components/src/components/common/select/select.tsx
ivy-edp: This is necessary because the select component automatically sizes itself based on the width of the selected item.
Without setting a width, it can end up looking like this:
grafik


By defining a width, I can ensure the dropdown menu is at least somewhat wider. However, if an item’s text is longer than the defined width, it will still get cut off:
grafik

packages/graph/src/data/data.ts
ivy-lli, line 1: this data.ts is, rather than the other two files, only used in the story right? so IMO it should not live here in this folder, but beside the story...

packages/graph/src/nodes/graphNode.css.ts
```diff
@@ -0,0 +1,69 @@
...
+  border: '2px solid var(--P500)'
+});
+
+export const originalNode = style({
+  border: '2px solid  var(--success-color)'
+});
+
+export const normalNode = style({
+  border: '1px solid  var(--body)'
+});
```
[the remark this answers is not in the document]
ivy-edp: Nice, i like it. Thanks 👍

packages/graph/src/data/getLayoutedElements.ts
```diff
@@ -0,0 +1,47 @@
+import type { GraphNode } from '../graph';
+import Dagre from '@dagrejs/dagre';
+import { type Edge } from '@xyflow/react';
+
+const dagreGraph = new Dagre.graphlib.Graph().setDefaultEdgeLabel(() => ({}));
+export type Direction = 'TB' | 'LR';
```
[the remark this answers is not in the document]
ivy-edp: unfortunatly not. This is just a string: 
grafik

packages/graph/src/data/getLayoutedElements.ts
```diff
@@ -0,0 +1,47 @@
...
+    if (node.measured) {
+      if (node.measured.height) {
+        nodeHeight = node.measured.height;
+      }
+      if (node.measured.width) {
+        nodeWidth = node.measured.width;
+      }
+    }
+    dagreGraph.setNode(node.id, { width: nodeWidth, height: nodeHeight });
+  });
```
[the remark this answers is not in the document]
ivy-edp: Did unforuntatly not work, because we reuse the same nodeSize object for all nodes, so when its width or height changed in the loop, it affected all nodes because they all referenced the same object. Or at least thats what i think is happening.

tsconfig.json
ivy-lli, line 1: seems still to have wrong encoding.

ivy-lli (approved):

tsconfig.json
ivy-edp: Strange... Before it was on CRLF and now on LF 🤔

ivy-lli: when it is now LF then it is good, however on my machine and branch it already is LF...

PR #369, opened 2025-03-06
XIVY-16149 Add new layout-icons

ivy-edp:
In the review, it was requested that it should be a little clearer what the individual ‘Align items’/‘Justify Content’ settings in the layout component mean, which is why we now want to add icons now.


ivy-edp:
Can i add these icons also to 12 or do i break something with that change?

ivy-lli:
not sure where you want to use them, icons looks good. and no should not break something. as long as you do not rename or remove old icons.
maybe rename them to aligh-left, align-right? so they are streamlined with the existing aling-horizontal/vertical?

ivy-lli (approved):

PR #376, opened 2025-03-14
XIVY-16222 Support table key arrow up/down scroll with virtualized rows

ivy-lli:
Support virtualized rows when scroll with arrow keys:

Screen Recording 2025-03-14 at 12 16 56
instead of:
Screen Recording 2025-03-14 at 12 20 13

### DIFF
--- a/packages/components/src/components/common/table/hooks/hooks.tsx
+++ b/packages/components/src/components/common/table/hooks/hooks.tsx
@@ -1,4 +1,5 @@
 import { SearchInput } from '@/components/common/input/input';
+import { ROW_VIRTUALIZE_INDEX_ATTRIBUTE } from '@/components/common/table/table';
 import { resetAndSetRowSelection, selectRow } from '@/utils/table/table';
 import {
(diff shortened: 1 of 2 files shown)


packages/components/src/components/common/table/hooks/hooks.tsx
```diff
@@ -200,7 +201,7 @@ export const useTableKeyHandler = <TData,>({ table, data, options }: TableKeyboa
       table.resetRowSelection();
       allRows[newReorderIndex].toggleSelected();
       setRootIndex(newReorderIndex);
-      event.currentTarget.rows[newReorderIndex].scrollIntoView({ block: 'center' });
```
ivy-lli, line 203: the problem here was that event.currentTarget.rows only contains the virtualized rows. So as soon as there are rows at the start which only exists in the memory but not as dom nodes the index is no longer matching.

The new solution checks if there is a data-vindex attribute on the row and tries to get the correct row via this attribute.

ivy-edp (approved):

PR #392, opened 2025-03-31
Add palette

ivy-lli:


ivy-edp (approved):
Very nice 👍 
One thought I had was that it might be cool to define a background color directly in the section or item config, alongside the icon. As this is required in the Process-Editor:
grafik

 Of course you can (if I have understood everything correctly) also achieve this if you create a CustomPaletteItem and use this, but this might be an advantage, as you can use the BasicPalette component.

ivy-lli:
> Very nice 👍 One thought I had was that it might be cool to define a background color directly in the section or item config, alongside the icon. As this is required in the Process-Editor: grafik
> 
> Of course you can (if I have understood everything correctly) also achieve this if you create a CustomPaletteItem and use this, but this might be an advantage, as you can use the BasicPalette component.

I think we can easily do this via CSS, but I add some more classes to make this easy possible 👍

PR #408, opened 2025-04-14
XIVY-16414 condition-builder support for long values

ivy-fhe:
- specifies for the condition-values to grow, but also shrink below content size
condition-builder-basic
condition-builder-isTrue


ivy-lli (approved):
Exaclty something like that, maybe the story has the flex: 1 defined? In this case you should no be able to remove it there.

PR #414, opened 2025-04-22
Move storybook to root

ivy-lli:
As more packages will follow also with components I think it makes sense to have the storybook in the root. 
E.g. after this we can easily put the graph stuff into an own package which also has stories which will (or should) showup in the storybook.


ivy-edp (approved):
👍

PR #424, opened 2025-04-29
XIVY-15944 include info when searching in browser

ivy-lgi:


ivy-lgi:
This might break tests in other places :3

packages/components/src/components/editor/browser/browser.tsx
```diff
@@ -63,6 +63,11 @@ export const useBrowser = (
   ];
 
   const [filter, setFilter] = React.useState(options?.initialSearch ?? '');
+  const globalFilterFn = (row: Row<BrowserNode>, _columnId: string, filterValue: string) => {
+    filterValue = filterValue.toLowerCase();
```
ivy-lli, line 67: It is noramlly better to not change a variable which is provided by a parameter, better create a new one.

ivy-lli (approved):
👍

PR #437, opened 2025-05-12
Add dev ts-config package

ivy-lli:
Hello my fellow TS contributors (@ivy-lmu, @ivy-lgi, @ivy-edp, @ivy-cst)

I've started with a shared ts-config package, that can be used on our repos for basic tsconfig stuff, by add `@axonivy/ts-config` devDependency and extend the tsconfig from the shared one e.g. `"extends": "@axonivy/ts-config/integration"`.
More or less it only replaces the root config folder in all repos (like in this one). 

I also decided to rework the tsconfig a bit and make it more strict. 
Why now and not after the streamlining? Because now we can roll out this shared config repo by repo and do changes that are now "wrong". Inspired by: https://www.totaltypescript.com/tsconfig-cheat-sheet

The main change now is that getting an object from an array via index returns the `type | undefined`. e.g.
```ts
const array = [0, 1, 2];
const element = array[0]; //number | undefined;
```
Before this code resulted in element was from type `number`. However this was also true for `array[5]` what is not really good. 
The sad part is that it will enforce some changes. E.g. check a length and afterwards accessing the value will still result in a maybe undefined type:
```ts
const array = [0, 1, 2];
if (array.length > 0) {
  const element = array[0]; //number | undefined; 
}
```
This is because ts cannot infer this array so it know the size of it at dev time. It is of corse also true because an array can change "any" time at runtime and this could cause an error. However many of this problems can be fixed by check the accessed object e.g. like:
```ts
const array = [0, 1, 2];
const element = array[0]; //number | undefined; 
if (element !== undefined) {
  element; //number; 
}
```

So I think this adjustment makes writing code sometimes a little bit harder, however I think in general it is true that an element from an array is maybe undefined and needs to be checked...

So what do you think?

Main change
from: https://github.com/axonivy/ui-components/pull/437/files#diff-1542b47836923af953e9f14d4db6843171edb79232ba834276394b4ed3035c63
to: https://github.com/axonivy/ui-components/pull/437/files#diff-2637596346f828da23c1466413134b87e37cf382b13910f73c5361713a7b865f


ivy-lgi:
I like it :+1: 
Seems like good practice to make sure that the element is not `undefined`.

ivy-lgi (approved):
:fire:

PR #439, opened 2025-05-14
Fix graph in dark mode

ivy-lli:
- Reuse color variables
- Colors of nodes, edge labels, etc.
- React flow menus

Screenshot 2025-05-15 at 07 58 58

### DIFF
--- a/packages/graph/src/graph.css.ts
+++ b/packages/graph/src/graph.css.ts
@@ -1,0 +1,60 @@
+import { createGlobalThemeContract, createGlobalTheme } from '@vanilla-extract/css';
+import { vars } from '@axonivy/ui-components';
+
+const reactFlow = createGlobalThemeContract({
+  edge: {
+    stroke: 'xy-edge-stroke'
+  },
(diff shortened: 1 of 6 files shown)


packages/graph/src/graph.css.ts
```diff
@@ -0,0 +1,97 @@
...
+  controls: {
+    background: vars.color.n25,
+    backgroundHover: vars.color.n100,
+    color: vars.color.body,
+    colorHover: vars.color.body,
+    borderColor: vars.color.n200
+  }
+});
+
+// Default dark variables
```
ivy-edp, line 62: Not sure why these comments are needed?

ivy-edp (approved):
Very nice, thanks 👍

packages/graph/src/graph.css.ts
ivy-lli: This are all the variables which are supported from react flow CSS. I didn't define all of them but I will remove the comment.

PR #448, opened 2025-05-19
XIVY-16736 Empty palette message

ivy-lli:


ivy-edp (approved):
Nice 👍 I think this should also be merged into 12. Actually, in 12, I believe the form editor doesn't use the ui-component-palette; instead, it has its own implementation ✌️

ivy-lli:
Yes the form-editor on R12 has it's own palette impl

PR #450, opened 2025-05-20
XIVY-16716 Improve layout of initial rendered graph

ivy-edp:
Bruno's feedback was that the initially rendered graph didn’t look very polished.
This was because I didn’t know the actual node sizes at the beginning — they only become available afterward. So initially, I set all nodes to a fixed size of 172 x 50. Naturally, this caused some nodes with longer text to be misaligned or poorly positioned.

Now, I'm trying to estimate a more accurate width for each node based on the label length, which works reasonably well. I'm also applying a similar method to calculate the width of the select menu.

### DIFF
--- a/packages/graph/src/graph.tsx
+++ b/packages/graph/src/graph.tsx
@@ -3,7 +3,7 @@ import './graph.css';
 import { IvyIcons } from '@axonivy/ui-icons';
 import { Background, ConnectionMode, Controls, MiniMap, Panel, ReactFlow, ReactFlowProvider, type Node } from '@xyflow/react';
 import { BasicSelect, Button, Flex } from '@axonivy/ui-components';
-import { useGraph } from './data/useGraph';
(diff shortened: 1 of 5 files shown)


packages/graph/src/graph.tsx
```diff
@@ -50,6 +50,17 @@ const GraphRoot = ({ graphNodes, options }: GraphProps) => {
     options
   });
 
+  const selectItems = [
+    { value: 'all', label: 'Show all' },
```
ivy-lli, line 54: this label needs to be translated... so it should be provided by the implementation. 
Fine for me if the default impl has this but in the end the whole filter needs to come from the implemator I think...

ivy-lli (approved):

packages/graph/src/graph.tsx
ivy-edp: Yes, that works for me. I’m planning to enhance the filter option from a simple flag to an object with the following structure:
grafik

PR #466, opened 2025-05-28
Revert "XIVY-16846 fix auto focus of browser search"

ivy-lgi:
This reverts commit e2c8684a74a9f78bc6da9909bfdc13a136a4b625.

Using the supposed workaround breaks the keyboard navigation in the browser because after every selection change in the table the search gets focused again.
Also, I tried to recreate this bug using minimal elements and it the autofocus of an element inside a deep dialog worked.
Therefore, I think there is a problem somewhere we might be able to fix properly.


ivy-lgi:
https://github.com/axonivy/ui-components/pull/464

PR #483, opened 2025-06-03
XIVY-16737 Add attachment icon

ivy-edp:
@ivy-lli fyi


ivy-lli:
Why add this to 13.1? isn't this for the accordion tab change?

ivy-edp:
> Why add this to 13.1? isn't this for the accordion tab change?

Yes, you are right, only needed for master 👍

PR #491, opened 2025-06-10
XIVY-16737 Add InscriptionTabs-Component

ivy-edp:


packages/components/src/components/common/inscription-tabs/inscriptionTabs.stories.tsx
```diff
@@ -0,0 +1,20 @@
...
+};
+
+export default meta;
+
+type Story = StoryObj<typeof BasicInscriptionTabs>;
+
+export const Default: Story = {
+  render: () => {
+    const [tab, setTab] = React.useState('General');
+    return <BasicInscriptionTabs value={tab} onChange={setTab} tabs={insciptionTabsData} />;
```
ivy-edp, line 18: I only export the BasicInscriptionTabs-Component. Not sure if i should also export the other components, similar to the other Browser-Tab-Component, so we could do something like this:
grafik


what do you think?

ivy-lli:
We already have a tab component and yes this is at least from the styling and also from the features a bit different. However, why is this a new component and not a variant of the existing tab component?

packages/components/src/components/common/state/state.css.ts
ivy-lli, line 1: why full file changed?

packages/components/src/components/common/tabs/tabs.tsx
```diff
@@ -28,4 +41,63 @@ const TabsContent = ({ className, ...props }: React.ComponentProps<typeof TabsPr
...
+              ? 'warning'
+              : undefined;
+
+          return (
+            <TabsTrigger
+              key={`${index}-${tab.id}`}
+              className={cn(inscriptionTabsTrigger, 'ui-inscription-tabs-trigger')}
+              data-message={state}
+              value={tab.id}
+              aria-label={tab.id}
```
ivy-lli, line 78: I think this should be the tab.name

packages/components/src/components/common/tabs/inscriptionTabs.css.ts
ivy-lli, line 1: I do not understand...
Many of this styles are already existing in the tabs.css.ts. There also exists a variant for small tabs, why is this inscription style not another variant?

packages/components/src/components/common/tabs/inscriptionTabs.css.ts
```diff
@@ -0,0 +1,142 @@
...
+          display: 'none'
+        }
+      }
+    }
+  }
+});
+
+globalStyle(`${inscriptionTabsTrigger} i`, {
+  fontSize: '16px'
+});
```
ivy-lli: do a normal style which you apply to the IvyIcon component

packages/components/src/components/common/tabs/tabs.tsx
```diff
@@ -28,4 +41,60 @@ const TabsContent = ({ className, ...props }: React.ComponentProps<typeof TabsPr
...
+  const defaultTab = tabs.length > 0 ? tabs[0].id : '';
+  return (
+    <Tabs className={cn(inscriptionTabsRoot, 'ui-inscription-tabs')} defaultValue={defaultTab} value={value} onValueChange={onChange}>
+      <TabsList className={cn(inscriptionTabsList, 'ui-inscription-tabs-list', `tabs-${tabs.length >= 5 ? 'many' : 'few'}`)}>
+        {tabs.map((tab, index) => {
+          const state = tab.state.messages?.find(message => message.variant === 'error')
+            ? 'error'
+            : tab.state.messages?.find(message => message.variant === 'warning')
+              ? 'warning'
+              : undefined;
```
ivy-lli, line 68: why is this needed?

packages/components/src/components/common/tabs/tabs.tsx
```diff
@@ -28,4 +41,60 @@ const TabsContent = ({ className, ...props }: React.ComponentProps<typeof TabsPr
...
+
+          return (
+            <TabsTrigger
+              key={`${index}-${tab.id}`}
+              className={cn(inscriptionTabsTrigger, 'ui-inscription-tabs-trigger')}
+              data-message={state}
+              value={tab.name}
+              aria-label={tab.name}
```
ivy-lli, line 76: only the title and the aria-label should be the tab.name, not the value. otherwise the value gets the translated name

packages/components/src/components/common/tabs/tabs.tsx
ivy-edp: not needed 👍

packages/components/src/components/common/tabs/inscriptionTabs.css.ts
ivy-edp: Acctually just found out that this is not nessacary as there already exists a style ivy-16, which sets the font-size to 16px👍

packages/components/src/components/common/tabs/tabs.css.ts
```diff
@@ -1,11 +1,15 @@
 import { vars } from '@/styles/theme.css';
-import { style } from '@vanilla-extract/css';
+import { createContainer, style } from '@vanilla-extract/css';
 import { recipe, type RecipeVariants } from '@vanilla-extract/recipes';
 
+export const tabsListContainer = createContainer();
+export const tabsTriggerContainer = createContainer();
```
ivy-lli, line 6: I think you don't have to export the list container.
And I would add the inscriptiontab styles also in here and not split it to a separate file. Then you also would not have to export the other container.

ivy-lli (approved):
Yes the styles feel big now, nevertheless, it makes somehow more sense to me.
And what you also can do now is to add this inscription variant to the Default tab story as choosable variant.

PR #523, opened 2025-07-10
XIVY-1606 fix scrolling on basicDialog

ivy-jh:
https://1ivy.atlassian.net/browse/XIVY-16606


ivy-jh:
the main issue was that the flex element was already defined in the basic dialog. this had broken the relationship to the child element in the content when another flex element was used directly in the content.

Furthermore, we have prepared the basic dialog so that it is scrollable by default. only the maxheight has to be specified here.

@ivy-lgi, i hope my my explanation isnt misleading.

packages/components/src/components/common/dialog/dialog.css.ts
```diff
@@ -54,7 +54,9 @@ export const content = style({
   padding: '20px',
   backgroundColor: vars.color.background,
   border: vars.border.basic,
-  display: 'grid',
+  maxHeight: '80vh',
+  display: 'flex',
```
ivy-lli, line 58: This change can break existing Dialogs.
Not against it, but I think you have to check every dialog if everything still works as expected.

Maybe it is easier to just change the maxHeight of the basic dialog content?

ivy-lli (approved):

PR #544, opened 2025-07-23
XIVY-16606 Rethink basic browser

ivy-lli:
- Remove `BasicBrowser` component
  Enforces wrong implementation, where to much logic lives outside of the `DialogContent`
- Remove `DialogContent` from `BasicDialogContent`
  Enforces wrong implementation, where to much logic lives outside of the `DialogContent`
- `BasicDialogContent` only contains wished structure of a `DialogContent`
- Add `BasicDialogHeader`
  If `BasicDialogContent` is not ideal to use
- Add `BasicDialogFooter`
  If `BasicDialogContent` is not ideal to use

### DIFF
--- a/packages/components/src/components/common/dialog/dialog.tsx
+++ b/packages/components/src/components/common/dialog/dialog.tsx
@@ -28,9 +28,9 @@ const DialogContent = ({ className, children, ...props }: React.ComponentProps<t
     <DialogOverlay />
     <DialogPrimitive.Content className={cn(content, className, 'ui-dialog-content')} {...props}>
       {children}
-      <DialogPrimitive.Close asChild>
+      <DialogClose asChild>
         <Button icon={IvyIcons.Close} className={contentClose} />
-      </DialogPrimitive.Close>
+      </DialogClose>
     </DialogPrimitive.Content>
   </DialogPortal>
 );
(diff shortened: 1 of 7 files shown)


packages/components/src/components/common/dialog/dialog.tsx
```diff
@@ -56,46 +56,49 @@ const DialogDescription = ({ className, ...props }: React.ComponentProps<typeof
...
 );
-BasicDialogContent.displayName = 'BasicDialogContent';
+BasicDialogHeader.displayName = 'BasicDialogHeader';
 
-type BasicDialogProps = React.ComponentProps<typeof DialogPrimitive.Root> & {
-  dialogTrigger?: React.ReactNode;
-  contentProps: BasicDialoContentProps;
+type BasicDialogFooterProps = {
+  cancel: React.ReactNode;
+  submit: React.ReactNode;
```
ivy-lgi, line 74: These buttons are no longer optional, which means you have to provide them when creating a dialog.
I am not against that. But I am wondering if it might be better if they were optional, and in case they are missing, the default buttons are used. I think for the cancel button, the default button will be used more often, while for the submit button, a custom button will be used more often, or at least its text will be changed.

If a dialog contains only a browser, both buttons need to be set to `null` as the browser will provide the buttons on its own.

However, I also think that this approach makes it clearer when looking at an implementation where these buttons are coming from since they are there explicitly.

All in all, I'm fine with this, just some thoughts that popped up. Maybe you notice whether it feels good the way it is now when implementing the dialogs.

PR #553, opened 2025-08-06
Use PNPM instead of NPM

ivy-lli:


package.json
```diff
@@ -8,34 +8,41 @@
...
     "lint:fix": "eslint --fix",
     "lint:inspect": "eslint --inspect-config",
     "dev": "storybook dev -p 6006 --no-open",
     "test": "vitest",
     "test:ci": "vitest run --reporter=default --reporter=junit --outputFile=report.xml",
+    "raise:version": "lerna version --allow-peer-dependencies-update --no-git-tag-version --no-push --ignore-scripts --exact --yes",
     "publish:next": "lerna publish --exact --canary --preid next --pre-dist-tag next --no-git-tag-version --no-push --ignore-scripts --yes"
   },
   "devDependencies": {
+    "@axonivy/eslint-config": "workspace:*",
```
ivy-lmu, line 30: why is this needed?

ivy-lmu (approved):
cool thx 👍

package.json
ivy-lli: Because otherwise the root does not know the eslint bin and cannot trigger the linting

ivy-lmu: ah sorry lint is obvious, but why is `"@axonivy/ui-components": "workspace:*",` declared here?

ivy-lli: because storybook is in the root and it needs the styles from it

